PR #7, opened 2015-10-15
fix(file): fixing work with Dockerfile.*

antonikonovalov:
now it's not work becaouse file != context
see more here https://docs.docker.com/reference/commandline/build/

after we can change docs ^)

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,19 +12,19 @@ import (
 )
 
 type Docker struct {
-	Storage  string   `json:"storage_driver"`
-	Registry string   `json:"registry"`
-	Insecure bool     `json:"insecure"`
-	Username string   `json:"username"`
-	Password string   `json:"password"`
-	Email    string   `json:"email"`
-	Auth     string   `json:"auth"`
-	Repo     string   `json:"repo"`
-	Tag      string   `json:"tag"`
-	File     string   `json:"file"`
+	Storage  string `json:"storage_driver"`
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -121,8 +123,14 @@ func main() {
 	trace(cmd)
 	cmd.Run()
 
+	strCmd := []string{"/usr/bin/docker", "build", "--pull=true","--rm=true", "-t", vargs.Repo, vargs.Context}
+	// Add file flag to cmd
+	if len(vargs.File) != 0 {
+		strCmd = append(append(strCmd[0:3],"-f="+vargs.File),strCmd[4:]...)
+	}
```
bradrydzewski: I'm having a bit of difficulty reading the above code with the slicing/appending. Do you think we could re-write like this:

```Go
if len(vargs.File) == 0 {
	vargs.File = "Dockerfile"
}
if len(vargs.Context) == 0 {
	vargs.Context = "."
}
cmd = exec.Command("/usr/bin/docker", "build", "--pull=true", "--rm=true", "-t", "-f", vargs.File, vargs.Repo, vargs.Context)
```

antonikonovalov: yes, it's simple :+1:

antonikonovalov: @bradrydzewski  done

bradrydzewski:
sorry, meant to merge this sooner ... getting pulled in many different directions these days :)

PR #9, opened 2015-10-21
fixes #8

### LINKED ISSUE #8: Plugin doesn't work with custom private registry
The problem lies here:
https://github.com/drone-plugins/drone-docker/blob/master/main.go#L41

I have custom private registry with urls like: `my.registry/image-name:tag`, so when I enter as `repo`:
- `image-name` I get `image-name:tag`
- `registry.host/image-name` I get `registry.host/registry.host/image-name:tag`

gregory90:


gregory90:
Thanks! Is image plugins/drone-docker on DockerHub auto updated?

PR #10, opened 2015-10-24
Fix `vars` typo

floatdrop:
`vars` -> `vargs`


bradrydzewski:
great catch. I realized I hadn't set this project up to auto-build yet. Will do that now

PR #13, opened 2015-11-09
Add support for saving/restoring image layers from archive

janeczku:
This PR adds support to the Docker plugin for saving/restoring Docker image layers to/from an archive file. Combined with the cache plugin it enables caching of image layers between builds.

Adds a new parameter:

##### `archive`
Create a tarred repository containing all layers of the Docker image and restore the image repository on consecutive builds.
The `archive` value must be a dictionary defined with a required `file` and an optional `tag` key.

`file` can either be an absolute path starting with `/drone` or a relative path which will be expanded from the working directory.
`tag` allows limiting the export of the image layers to the supplied tag(s).

```yaml
archive:
  file: docker/image.tar
  tag: latest
```

##### Example usage:

```yaml
publish:
  docker:
    username: kevinbacon
    password: $$DOCKER_PASSWORD
    email: kevin.bacon@mail.com
    repo: foo/bar
    tag:
      - latest
      - "1.0.1"
    archive:
      file: docker/image.tar
      tag: latest

cache:
  mount:
    - docker/image.tar
```

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -9,6 +9,9 @@ The following parameters are used to configure this plugin:
 * `tag` - repository tag for the image
 * `insecure` - enable insecure communication to this registry
 * `storage_driver` - use `aufs`, `devicemapper`, `btrfs` or `overlay` driver
+* `archive` - save and restore image layers to/from a tarred archive
+    * `file` - absolute or relative path to archive file
+    * `tag` - limit archiving to these tag(s) (optional)
 
 The following is a sample Docker configuration in your .drone.yml file:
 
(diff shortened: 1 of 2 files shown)


bradrydzewski:
awesome, I really like this proposal and the code looks great. One minor comment that I have is that it would be nice to be able to save and load from anywhere inside `/drone`. If we join the `workspace.Path` it will force saving and loading from inside `/drone/src/github.com/{owner}/{name}/`

I'm thinking instead of doing `filepath.Join`:

```
tarFile := filepath.Join(workspace.Path, vargs.Archive)
```

we can maybe use `filepath.Rel` to get the relative path to workspace? and then `stat` the relative path? what do you think?

also we should add the new archive parameter to the `DOCS.md` file :)

bradrydzewski:
maybe something like this? it would allow an absolute path or relative path to the workspace:

```
	tarFile, err := filepath.Rel(workspace.Path, vargs.Archive)
	if err != nil {
		tarFile = filepath.Join(workspace.Path, vargs.Archive)	
	}
```

janeczku:
Thats a great suggestion!  :+1: 
I will update the PR later today with the code change and the documentation.

bradrydzewski:
thanks!

PR #14, opened 2015-11-19
Separate 'archive' parameter in 'load' and 'save'

janeczku:
This is a follow up to PR #13 allowing the user to `load` and `save` image layers from/to separate files.

```yaml
load: docker/image.tar
save:
  file: docker/image.tar
  tags: latest
```

@bradrydzewski


bradrydzewski:
awesome, thanks!

PR #15, opened 2015-11-22
Adding support to registry mirroring

atosatto:
It is now possibile to optionally specify a mirror registry instead of pulling images directly from the central Hub.


atosatto:
Thank you! :smiley:

PR #20, opened 2015-12-03
docker image update

ipedrazas:
update from rancher/docker:1.8.1 to 1.9.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 #
 #     docker build --rm=true -t plugins/drone-docker .
 
-FROM rancher/docker:1.8.1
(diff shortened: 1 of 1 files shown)


msteinert:
lgtm

@bradrydzewski, Is there a specific reason for using the `rancher/docker` image rather than [official docker image][]?

[official docker image]: https://hub.docker.com/_/docker/

bradrydzewski:
@msteinert yes, the rancher docker image uses scratch as the base image, and is therefore only around 20mb in size, which reduces plugin download time

bradrydzewski:
LGTM as well

PR #24, opened 2015-12-27
add debian wheezy/jessie kernel workaround

xinity:


tboerger:
This PR is outdated, maybe rebase?

xinity:
so it seems , you guys took your time to review it :)

should be right now, let me know

DOCS.md
```diff
@@ -20,7 +20,6 @@ The following parameters are used to configure this plugin:
     * `destination` - absolute / relative destination path
     * `tag` - cherry-pick tags to save (optional)
 * `load` - restore image layers from the specified tar file
-* `build_args` - [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables-build-arg) to pass to `docker build`
```
tboerger: This should not be dropped.

DOCS.md
```diff
@@ -28,7 +27,7 @@ The following is a sample Docker configuration in your .drone.yml file:
 publish:
   docker:
     username: kevinbacon
-    password: pa55word
+    password: $$DOCKER_PASSWORD
```
tboerger: Why did you replace that?

DOCS.md
```diff
@@ -42,7 +41,7 @@ You may want to dynamically tag your image. Use the `$$BRANCH`, `$$COMMIT` and `
...
```
tboerger: Why did you replace that?

DOCS.md
```diff
@@ -54,7 +53,7 @@ Or you may prefer to build an image with multiple tags:
...
```
tboerger: Why did you replace that?

DOCS.md
```diff
@@ -64,19 +63,6 @@ publish:
 ```
 
 Note that in the above example we quote the version numbers. If the yaml parser interprets the value as a number it will cause a parsing error.
-
```
tboerger: This should not be dropped.

DOCS.md
```diff
@@ -86,7 +72,7 @@ The Drone build environment is, by default, ephemeral meaning that you layers ar
 publish:
   docker:
     username: kevinbacon
-    password: pa55word
+    password: $$DOCKER_PASSWORD
```
tboerger: Why did you replace that?

tboerger:
Please revert your wrong drops, than you will get my approval :)

DOCS.md
gtaylor: FWIW, I like this change. It is consistent with best practices.

donny-dont: Should there maybe be a section offering suggestions on what should be secured? The $$ is best practices but might also make things less clear.

gtaylor: I could see sneaking a link to the secrets docs up in the param list, or in a comment in the yaml example itself.

tboerger:
@xinity any further fixes? :)

let4be:
I wish I saw it in the beginning, spent 3 hours trying to find this solution :(
Why nobody checks such things...

DOCS.md
bradrydzewski: I vote to not use the `$$` notation in the documentation. It is likely we could change this notation in a future release, or make secrets pluggable, in which case we might not want the notation hard-coded into all the docs. 

I would recommend we instead add a section to the plugin template (so it appears on the all the plugin docs) that recommends using secrets instead of hard-coding into the yaml, and provides a link to the secrets documentation.

tboerger:
Replaced by #41

PR #25, opened 2016-01-03
Docker graph caching

djmaze:
This allows caching the whole Docker graph with the cache plugin. Caching is efficient with the overlay storage driver at least. Works much better (also faster) than the existing save/load approach for me. 

The drawback with this approach: old image layers are preserved in the graph directory belonging to the repository. This is why I added code which removes untagged images after the image has been pushed. This makes sure the cache doesn't grow over time. (We could add a configuration option for this, as it is only needed for this kind of caching.)

This caching method probably shouldn't be used with other storage drivers. I only tried `overlay` (good) and `devicemapper` (totally inefficient). Not quite sure if it works with `aufs` (probably only workable if host system also uses aufs). Anyone able to try?

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -73,10 +73,36 @@ publish:
     build_args:
       - HTTP_PROXY=http://yourproxy.com
 ```
- 
-## Layer Caching
 
-The Drone build environment is, by default, ephemeral meaning that you layers are not saved between builds. The below example combines Drone's caching feature and Docker's `save` and `load` capabilities to cache and restore image layers between builds:
+## Caching
+
+The Drone build environment is, by default, ephemeral meaning that you layers are not saved between builds. There are two methods for caching layers.
+
+### Graph directory caching
+
+This is the preferred method when using the `overlay` storage driver. Just use Drone's caching feature to backup and restore the directory `/drone/docker`, as shown in the following example:
+
+```yaml
+publish:
+  docker:
+    username: kevinbacon
+    password: pa55word
+    email: kevin.bacon@mail.com
+    repo: foo/bar
+    tag:
+      - latest
+      - "1.0.1"
+
+cache:
+  mount:
+    - /drone/docker
+```
+
+NOTE: This probably won't work correctly with the `btrfs` driver, and it will be very inefficient with the `devicemapper` driver. Please make sure to use the `overlay` storage driver with this method.
+
+### Layer Caching
+
+The below example combines Drone's caching feature and Docker's `save` and `load` capabilities to cache and restore image layers between builds:
 
 ```yaml
 publish:
--- a/DOCS.md
+++ b/DOCS.md
@@ -76,7 +76,7 @@ publish:
 
 ## Caching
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -222,6 +223,29 @@ func main() {
 		}
 	}
 
+	// Remove untagged images, if any
+	var outbuf bytes.Buffer
+	cmd = exec.Command("sh", "-c", "docker images | grep '^<none>' | awk '{print $3}'")
```
tboerger: docker images -q -f dangling=true
should be much better.

djmaze: Thanks! Didn't remember that one. Changed.

DOCS.md
```diff
@@ -73,10 +73,36 @@ publish:
     build_args:
       - HTTP_PROXY=http://yourproxy.com
 ```
- 
-## Layer Caching
 
-The Drone build environment is, by default, ephemeral meaning that you layers are not saved between builds. The below example combines Drone's caching feature and Docker's `save` and `load` capabilities to cache and restore image layers between builds:
+## Caching
+
+The Drone build environment is, by default, ephemeral meaning that you layers are not saved between builds. There are two methods for caching layers.
```
tboerger: `your layers`

main.go
```diff
@@ -222,6 +223,29 @@ func main() {
...
+	cmd.Stderr = os.Stderr
+	trace(cmd)
+	err = cmd.Run()
+	if err != nil {
+		os.Exit(1)
+	}
+
+	if outbuf.Len() > 0 {
+		images := strings.Split(strings.TrimSpace(outbuf.String()), "\n")
+		cmd = exec.Command("docker", append([]string{"rmi"}, images...)...)
```
tboerger: What about `cmd = exec.Command("docker", "rmi", images...)`?

tboerger:
I'm not sure if @bradrydzewski likes this change but codewise LGTM

main.go
djmaze: Tried that. It doesn't seem to be possible to add a single parameters as well as an array slice for a varargs parameter (`too many arguments in call to exec.Command`). 

Not a Go expert though.

bradrydzewski:
@djmaze if I create a `develop` branch are you open to pushing this PR to that branch? For some of the heavily used plugins (docker, git, etc) I'm thinking of beta-testing changes in a develop branch before publishing to master. The yaml supports tags so it would work like this:

```yml
publish:
  docker:develop:
    ...
```

djmaze:
@bradrydzewski Yes, why not? That sounds like a good idea.

studna:
Nice feature, +1.

djmaze:
Just tested it with the `aufs` storage driver. (AUFS is automatically used inside the drone-docker container without configuring.) 

Seems to work fine as well. I adjusted the documentation accordingly.

tboerger:
LGTM

bradrydzewski:
@djmaze we have the [develop](https://github.com/drone-plugins/drone-docker/tree/develop) branch ready if you can re-open the PR against that branch. We can then merge right away. Thanks!

djmaze:
Okay, done!

PR #40, opened 2016-02-13
Fix docker d

casualjim:
the argument -d doesn't seem to work anymore

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -27,6 +27,7 @@ The following is a sample Docker configuration in your .drone.yml file:
 ```yaml
 publish:
   docker:
+    privileged: true
     username: kevinbacon
     password: pa55word
     email: kevin.bacon@mail.com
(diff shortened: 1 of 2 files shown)


DOCS.md
```diff
@@ -27,6 +27,7 @@ The following is a sample Docker configuration in your .drone.yml file:
 ```yaml
 publish:
   docker:
+    privileged: true
```
bradrydzewski: Explicitly setting privileged mode for this plugin isn't required since this plugin is automatically whitelisted to run in privileged mode

bradrydzewski:
This plugin runs docker 1.9 and works with -d ... we use it to publish drone and all the drone plugins. Is there maybe a different issue / root cause?

casualjim:
it didn't want to work for me until I made these 2 changes and flagged the project as trusted (not in that order)

tboerger:
Please revert the changes to the docs, than we can merge the `daemon` instead of `-d` change.

casualjim:
I've reverted the docs update

jackspirou:
@casualjim looks like the branch is now out of date

casualjim:
sorry about that, i rebased it

bradrydzewski:
this change exists in the `develop` branch and is published to dockerhub, so you can test with the following change to your yaml:

```diff
publish:
- docker:
+ docker:develop:
```

once we've had some time to test out the develop branch we'll merge to master

Issue #43, opened 2016-02-27
Use commit tag name for tagging an image


grahovam:
So far, an image can be variably tagged with $$BRANCH, $$COMMIT or $$BUILD_NUMBER. It is not possible to use a commit tag, right? So that if I push a tag on my git repo, a docker build is being triggered with an image tag of the same tag name?

bradrydzewski:
yep, you can use $$TAG ... you can also do `$${TAG##v}` to remove the `v` prefix on your tag name. You need to enable builds for tags, which can be done in your drone repository setting screen.

feel free to ask these sort of questions in the gitter room at https://gitter.im/drone/drone

PR #44, opened 2016-02-28
Update DOCS.md corresponding to #43

grahovam:
Update DOCS.md corresponding to #43

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -49,7 +49,7 @@ publish:
 ```
 
 If you use the `$$TAG` variable, you need to enable builds for tags, which can be done in your drone repository setting screen.
-You can also do `$${TAG##v}` to remove the v prefix on your tag name.
+You can also do `$${TAG##v}` to remove the `v` prefix on your tag name.
 
 Or you may prefer to build an image with multiple tags:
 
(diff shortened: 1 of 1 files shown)


DOCS.md
```diff
@@ -48,6 +48,9 @@ publish:
     tag: $$BRANCH
 ```
 
+If you use the `$$TAG` variable, you need to enable builds for tags, which can be done in your drone repository setting screen.
+You can also do `$${TAG##v}` to remove the `v` prefix on your tag name.
```
tboerger: I would prefer a link to the docs for the general string substitution

tboerger:
This documentation also matches for drone-ecr and drone-gcr

DOCS.md
grahovam: Where can I find those?

bradrydzewski:
I'm closing this because it is documented globally for all plugins at http://readme.drone.io/usage/variables/#string-interpolation:2b8b8ac4006be88c769f5e3fd99b009a

@tboerger I don't want to put the direct link in the docs either for a few reasons. If we put a directly link here, then we need to put the link in the docs for the other 50 plugins. And then we should probably include a link to the .drone.sec file documentation for all 50 plugins. And then think how annoying it will be when we change one of those URLs and have to update 50+ files :smile:  

If anything, we should put links in the master template so that it should up automatically for all of the plugins.

PR #46, opened 2016-03-15
Add an option to pull images before builds

grampajoe:
As an alternative to caching images on Drone workers, this adds the option to pull an image from a registry before building. If you pull an earlier version of the image you're building, it can act as a remote cache to avoid relying on Drone's local caching.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -20,6 +20,9 @@ The following parameters are used to configure this plugin:
     * `destination` - absolute / relative destination path
     * `tag` - cherry-pick tags to save (optional)
 * `load` - restore image layers from the specified tar file
+* `pull_before` - pull an image before building
+    * `repo` - repository of the image to pull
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Hey there! In this case `pull` is a reserved keyword in the yaml used to pull the latest version of the image or plugin. It is primarily used by build steps, however, build and plugin steps share the same data structure:

```
build:
  image: golang
  pull: true
```

I also wonder if this would throw an error trying to parse the yaml, before the build begins, since it is expecting a boolean value.

grampajoe:
Hi @bradrydzewski, you're right! I changed it to `pull_before`, and I'm in the process of giving it a more thorough integration test.

bradrydzewski:
Great thanks! Also could you send the PR to the develop branch? We use the develop branch to incubate changes before merging

grampajoe:
Whoops! Will do.

PR #53, opened 2016-05-02
Support branch name as tag

mujiburger:


bradrydzewski:
this is already supported at the yaml level using string interpolation. See http://readme.drone.io/usage/variables/#string-operations:2b8b8ac4006be88c769f5e3fd99b009a

you can do the following:

```
publish:
  docker:
    tag: $$BRANCH
```

PR #55, opened 2016-05-02
Move away from drone-plugin-go and fix the example in README

mujiburger:


README.md
```diff
@@ -35,15 +35,15 @@ docker run -i --privileged -v $(pwd):/drone/src plugins/drone-docker <<EOF
...
+        "author_email": "john.smith@gmail.com",
         "event": "push",
         "branch": "master",
         "commit": "436b7a6e2abaddfd35740527353e78a227ddcb2c",
         "ref": "refs/heads/master"
     },
     "workspace": {
         "root": "/drone/src",
-        "path": "/drone/src/github.com/drone/drone"
+        "path": "/drone/src"
```
tboerger: Wrong, root is just this, path should stay like it was

mujiburger:
Please check out the update.   The problem is that workspace path is used within the plugin.  So the example as provided throws errors because the path doesn't exist.   Hopefully using this plugin's realistic path and mounting . to the appropriate volume provides a realistic and functional example.

tboerger:
Looks better now.

bradrydzewski:
closing since these changes are based on 0.4 plugins and we updated master last week to 0.5, to which these changes no longer apply

PR #61, opened 2016-06-18
More logging, disable docker login email, changes to makefile

vespakoen:
First go code I ever wrote (copy and pasted actually) so check carefully =)

- Add codegangsta/cli to the deps (what are the 3 dots at the end for? e.g.: ...)
- Add more logging (this seems to be logged later on as well by `docker info` perhaps, will remove)
- Change IMAGE / tag in Makefile (so it can be picked up by drone-agent)
- Remove email from docker login (is deprecated https://github.com/docker/docker/blob/master/docs/deprecated.md#-e-and---email-flags-on-docker-login)


Makefile
```diff
@@ -13,6 +13,7 @@ clean:
 	go clean -i ./...
 
 deps:
+	go get -u github.com/codegangsta/cli/...
 	go get -t ./...
```
bradrydzewski: `go get -u github.com/codegangsta/cli/...` is not required. You should be building with `GOVENDOREXPERIMENT=1` which uses the `/vendor` directory for dependencies

vespakoen:
I copied the build config from here: https://github.com/drone/drone/blob/master/.drone.yml#L19
Perhaps it should be moved to the `commands` section?

plugin.go
```diff
@@ -95,11 +95,11 @@ func (p Plugin) Exec() error {
 	cmds = append(cmds, commandVersion())      // docker version
 	cmds = append(cmds, commandInfo())         // docker info
 	cmds = append(cmds, commandBuild(p.Build)) // docker build
-
 	for _, tag := range p.Build.Tags {
+		fmt.Fprintf(os.Stdout, "Tagging: %s\n", tag)
 		cmds = append(cmds, commandTag(p.Build, tag)) // docker tag
-
 		if p.Dryrun == false {
+			fmt.Fprintf(os.Stdout, "Pushing: %s\n", tag)
```
bradrydzewski: this logging is un-necessary because we trace every command that gets executed with the trace function:

```
func trace(cmd *exec.Cmd) {
	fmt.Fprintf(os.Stdout, "+ %s\n", strings.Join(cmd.Args, " "))
}
```

as a result, you see this in your drone logs:

```
+ /usr/local/bin/docker tag 2530f8723f6c960d1ba389db4acaf5fe69d8550d drone/drone:0.5.0
```

bradrydzewski:
> I copied the build config from here: https://github.com/drone/drone/blob/master/.drone.yml#L19
Perhaps it should be moved to the commands section?


the use of `GO15VENDOREXPERIMENT=1` is part of the `environment` section of the yaml and it is also discussed in the README

vespakoen:
That leaves only this:

https://github.com/drone-plugins/drone-docker/pull/61/files

bradrydzewski:
Before we can merge we need @donny-dont to verify that he can connect to his corporate registry without email address, since older registries may still have this requirement.

We are very conservative with accepting pull requests for this plugin because it is used by thousands of teams and we need to make sure we don't break existing Drone installations.

bradrydzewski:
what about something like this?

```diff
function commandLogin(p.Login) *exec.Cmd {
+	if p.Login.Email != "" {
+		return commandLoginEmail(p.Login)
+	}
	return exec.Command(
		dockerExe, "login",
		"-u", login.Username,
		"-p", login.Password,
-		"-e", login.Email,
		login.Registry,
	)
}

+function  commandLoginEmail(p.Login) *exec.Cmd {
+	return exec.Command(
+		dockerExe, "login",
+		"-u", login.Username,
+		"-p", login.Password,
+		"-e", login.Email,
+		login.Registry,
+	)
+}
```

vespakoen:
Brought that suggestion in

plugin.go
```diff
@@ -123,6 +123,18 @@ const dockerExe = "/usr/local/bin/docker"
 
 // helper function to create the docker login command.
 func commandLogin(login Login) *exec.Cmd {
+	if p.Login.Email != "" {
+		return commandLoginEmail(p.Login)
+	}
```
bradrydzewski: sorry my pseudo code was off :)

```diff
-return commandLoginEmail(p.Login)
+return commandLoginEmail(login)
```

plugin.go
```diff
@@ -123,6 +123,18 @@ const dockerExe = "/usr/local/bin/docker"
...
+	return exec.Command(
+		dockerExe, "login",
+		"-u", login.Username,
+		"-p", login.Password,
+		login.Registry,
+	)
+}
+
+func commandLoginEmail(p.Login) *exec.Cmd {
 	return exec.Command(
```
bradrydzewski:
```diff
-func commandLoginEmail(p.Login) *exec.Cmd {
+func commandLoginEmail(login Login) *exec.Cmd {
```

bradrydzewski:
sorry my pseudo code was off and since the output at beta.drone.io is busted, this was what caused the failure:

```
+ go test -cover -coverprofile=coverage.out
# github.com/drone-plugins/drone-docker
./plugin.go:126: undefined: p in p.Login
./plugin.go:127: undefined: p in p.Login
./plugin.go:137: undefined: p in p.Login
./plugin.go:140: undefined: login in login.Username
./plugin.go:141: undefined: login in login.Password
./plugin.go:142: undefined: login in login.Email
./plugin.go:143: undefined: login in login.Registry
```

I can see the logs correctly on my local update of the UI :)

bradrydzewski:
still seeing compile errors -- make sure you are building locally to catch these :)

```
+ go test -cover -coverprofile=coverage.out
# github.com/drone-plugins/drone-docker
./plugin.go:126: undefined: p in p.Login
```

bradrydzewski:
awesome, thanks so much for your patience and working with me on this!

bradrydzewski:
LGTM

vespakoen:
Thank **you** for being so patient with me ;)

So the IMAGE change in the `Makefile` is ok as well?

PR #67, opened 2016-07-22
Streamline, proper documentation, cleanup

tboerger:


bradrydzewski:
also do we _need_ the vendor.yml file? one less file in the root would be nice, unless there is a compelling reason to start adding it. I've been omitting them

.drone.yml
```diff
@@ -1,21 +1,35 @@
...
+      # - go vet
       - go test -cover -coverprofile=coverage.out
-      - go build -ldflags "-s -w -X main.build=$DRONE_BUILD_NUMBER"
-  docker:
-    storage_driver: vfs
+      - go build -ldflags "-s -w -X main.build=$DRONE_BUILD_NUMBER" -a -tags netgo
+
+  latest:
+    image: docker
+    storage_driver: overlay
```
bradrydzewski: we no longer need to set this, in fact we should omit, because the docker plugin will choose the best driver at runtime

README.md
```diff
@@ -1,18 +1,21 @@
...
 Build the binary with the following commands:
 
 ```
-export GO15VENDOREXPERIMENT=1
-export GOOS=linux
-export GOARCH=amd64
-export CGO_ENABLED=0
-
-go build -a -tags netgo
+GO15VENDOREXPERIMENT=1 CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -tags netgo
```
bradrydzewski: we can omit GO15VENDOREXPERIMENT from the docs because Go 1.6 doesn't require the flag

also I don't think we should set the GOOS or GOARCH or -netgo tags here since someone might be testing the raw binary locally. Instead maybe this should be added to the docker build step

bradrydzewski: for an example see https://github.com/drone-plugins/drone-git#docker which separates the regular build process from the docker build process

.drone.yml
```diff
@@ -1,21 +1,35 @@
...
     repo: plugins/docker
-    tag: [ "latest", "1.0", "1" ]
+    tags: [ "latest", "1.0", "1" ]
     when:
       branch: master
       event: push
 
+  develop:
+    image: docker
```
bradrydzewski: This can be removed. I'm not doing `develop` tags anymore. This has been replaced with plugin versioning (above)

tboerger: Do we also want to post different becomes per build, e.g. put the build number into it?

bradrydzewski: not at this time. official plugins will promote stability over features, and (in theory) will not have many updates

tboerger:
@bradrydzewski I have applied all comments from the streamline PRs.

Bugagazavr:
LGTM

donny-dont:
LGTM

PR #76, opened 2016-08-26
fixed appending registry before the repo

Zhomart:
It should prepend `p.Login.Registry` only if login.registry is not equal to defaultRegistry and build.repo is not started with login.registry.


bradrydzewski:
thanks!

Zhomart:
This is not good solution either. Because registry can be defined in multiple ways, like "https://gcr.io" or "gcr.io", or it could be something else. 

I'll send another PR.

bradrydzewski:
I think the docker daemon will accept both. Docker will first try to use https and then try to use http. So I'm not sure any changes are needed unless there are specific errors that are going to be encountered?

Zhomart:
It creates an error, I fixed it.

PR #77, opened 2016-08-26
fixed appending registry before the repo smart

Zhomart:
It checks for prefix after removing protocols.

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 	"time"
 )
@@ -12,6 +13,10 @@ import (
 const (
 	// default docker registry
 	defaultRegistry = "https://index.docker.io/v1/"
+)
+
+var (
+	removeProtocolRegex = regexp.MustCompile(`^(\w+)://`)
 )
 
 type (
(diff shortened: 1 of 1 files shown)


tboerger:
Don't use regenerated, just use net/url :)

bradrydzewski:
could we just document this behavior that when using a custom registry, one should not include `http://` or `https://` in the registry url? I would prefer not to auto-correct invalid input. It can sometimes be more confusing that throwing an exception

bradrydzewski:
so I'd vote for killing this whole block

```diff
-	// this code attempts to normalize the repository name by appending the fully
-	// qualified registry name if otherwise omitted.
-	if p.Login.Registry != defaultRegistry &&
-		!strings.HasPrefix(p.Build.Repo, p.Login.Registry) {
-		p.Build.Repo = p.Login.Registry + "/" + p.Build.Repo
-	}
```

and adding a note to the documentation like this:

Please note that when using a custom registry (other than DockerHub) you will need to provide the registry URL and you will need to use a fully qualified repository name. For example:

```
plugin:
  docker:
    registry: http://registry.company.com
    repo: registry.company.com/my/image
```


... the reason is that when you tag an image locally you have to use `registry.company.com/my/image`. So providing this auto-concatination is inconsistent with how we use Docker locally and seems to cause some confusion. Best to keep it simple and remove the magic, as opposed to add to it.

Zhomart:
It makes sense.

PR #90, opened 2016-10-31
Adding proxy values to build args

donny-dont:
This makes it so the plugin implicitly adds proxy values as build args. It is done implicitly as the proxy values are contained within the agent and the information might not be known to the person creating the build.

Currently it checks to see if the values were set as build args and it will ignore the agent's values.

Also it sets the upper and lower case versions of the proxy values as I've found that this needs to happen.

@bradrydzewski @tboerger @Bugagazavr

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -27,7 +27,7 @@ type (
 		Bip           string   // Docker daemon network bridge IP address
 		DNS           []string // Docker daemon dns server
 		MTU           string   // Docker daemon mtu setting
-		IPv6	      bool     // Docker daemon IPv6 networking
+		IPv6          bool     // Docker daemon IPv6 networking
 	}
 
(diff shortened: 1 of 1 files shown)


plugin.go
```diff
@@ -108,6 +108,11 @@ func (p Plugin) Exec() error {
 		fmt.Println("Registry credentials not provided. Guest mode enabled.")
 	}
 
+	// add proxy build args
+	addProxyBuildArgs(&p.Build)
+
+	fmt.Printf("Build arg length %d", len(p.Build.Args))
```
tboerger: Do you want to keep that debug info?

tboerger:
LGTM

tboerger:
Now it's really fine for me

Bugagazavr:
LGTM

PR #103, opened 2017-01-27
Use dockerd instead of docker daemon

tuxity:
`docker daemon` is deprecated is 1.13, we should use new binary instead: `dockerd` (see https://github.com/docker/docker/pull/26834)

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -147,6 +147,7 @@ func (p Plugin) Exec() error {
 }
 
 const dockerExe = "/usr/local/bin/docker"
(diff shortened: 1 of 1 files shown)


tboerger:
It will be dropped with 1.16, so no hurry

tuxity:
Yep, but it doesn't hurt to switch it now

bradrydzewski:
LGTM

bradrydzewski:
> It will be dropped with 1.16, so no hurry

@tboerger very true, although now that we have proper versioning in place for this plugin, these sort of changes are a bit less risky than before :)

PR #105, opened 2017-01-27
Add verbose parameter

tuxity:
I've added a verbose parameter to do not run `docker version` and `docker info` command when not needed

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,11 @@ func main() {
 			Usage:  "dry run disables docker push",
 			EnvVar: "PLUGIN_DRY_RUN",
 		},
+		cli.BoolFlag{
+			Name:   "verbose",
+			Usage:  "output docker version and docker info comands",
+			EnvVar: "PLUGIN_VERBOSE",
+		},
 		cli.StringFlag{
 			Name:   "commit.sha",
 			Usage:  "git commit sha",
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -28,6 +28,11 @@ func main() {
 			Usage:  "dry run disables docker push",
 			EnvVar: "PLUGIN_DRY_RUN",
 		},
+		cli.BoolFlag{
+			Name:   "verbose",
+			Usage:  "output docker version and docker info comands",
+			EnvVar: "PLUGIN_VERBOSE",
```
tuxity: maybe I should add `Value:  "true",` to enable it per defaut ?

tboerger:
Why do you want to drop this information? It's useful to see that instead of adding another flag for that.

tuxity:
Because I don't really need it for my team, so thought it could be good to add a flag for that. But we can enable it per default

bradrydzewski:
@tuxity this is enabled by default to reduce support burden.  I'm not sure there is a lot of value to the broader community in making this configurable, so at this time my recommendation would be to fork the plugin (forking is highly encouraged) if you need custom output

bradrydzewski:
I should also point out that I think this can be handle with UI improvements, such as https://github.com/drone/drone-ui/issues/43

PR #107, opened 2017-02-02
Add ability to accept multiple push registries

greysond:
Like so:
```yaml
[...]
push_registries:
  - repo: something/else
  - registry: gcr.io/
    repo: a/b
  - registry: gcr.io/
    repo: a/c
  - registry: https://docker.lumerico.com/
    repo: super/secret
    username: roadhog
    password: timetogowholehog
```

Notes:
- If a registry is not specified, it inherits from the main config.
- Attempts to login to each registry before building. If the registry matches the main config and either the username is blank or the same as the main username, it skips login.
- Attempts to push each tag to each repo specified.

Thanks in advance for the review!

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


donny-dont:
I honestly think this makes things more confusing to users. Is there a reason why you don't just fork what's there?

bradrydzewski:
Agreed with @donny-dont. The small number of official plugins focus on simplicity and stability. Forking plugins or creating alternate versions that add more power or features is highly encouraged.

greysond:
That makes sense, thanks guys!

Issue #34, opened 2016-01-29
Drone doesn't seem to cache layers 


emilebosch:
I'm using this setup but i still get no cache performance increase. I do however see that is loading and saving however it doesn't seem to use the cache layers. Any idea? A normal build does use the cache layer. 

Is there a specific reason why we just don't mount the docker socket in the container and use the existing docker for building?

```
build:
  image: alpine:3.2

publish:
  docker:
    registry: XX
    username: XX
    password: $$REGISTRY_PASSWORD
    email: kevin.bacon@mail.com
    repo: XX
    load: docker/image.tar
    save:
      destination: docker/image.tar
      tag: latest

notify:
  slack:
    webhook_url: XX
    channel: update
    username: drone

cache:
  mount:
    - docker/image.tar
```

bradrydzewski:
I can't speak to the cache issue since this isn't something I've used personally, but perhaps @janeczku can comment since he was the one that introduced the feature.

> Is there a specific reason why we just don't mount the docker socket in the container and use the existing docker for building?

There are a few good reasons why we don't do this, but the most compelling is security. Mounting the host machine's Docker socket into the container would give your build environment root access to the host machine. This would make public facing Drone instances (such as the drone instance used to build this repository) vulnerable to a wide range of attacks.

janeczku:
@emilebosch Could you run your build with Docker debug output? See https://github.com/drone-plugins/drone-docker/blob/master/DOCS.md#troubleshooting

emilebosch:
@janeczku I've heard this security issue before. Although i think it doesn't really is too restricting. I understand the need, but i think allowing a configuration option in drone itself to "allow-unsafe-plugins" would mitigate this. Now, the drone decides from themselves what is considered "good" and "safe" which is ok practice. But severely limits my options. I would love to have direct access to the cluster it runs on so i can also schedule containers do blue green deployments etc. Since we run all the drone behind firewall it really would be nice for us.

bradrydzewski:
There is no reason you _have to_ use the docker plugin. You can mount the Docker daemon into your build container using volumes [1] and then run `docker build` and `docker push` from the build section. The plugin is all about safety, but drone won't stop you from doing almost exactly what you've described

[1] http://readme.drone.io/usage/build_test/#volumes:fb92aa3346185c57f15afda861d465a3

emilebosch:
[...]

PR #110, opened 2017-02-20
allowing docker to use cached base image in build

JeffDownie:
This fixes the last comment in issue https://github.com/drone-plugins/drone-docker/issues/34 - and adds a flag to allow _not_ pulling the base image down on every build.

I'm not 100% sure that the name I've given to the option is good, but I couldn't get the EnvVar options to work with a bool flag that was true-by-default correctly :/. This being my first go code ever, I'm happy to take any suggestions/help regarding this!

example usage:
```
  publish:
    image: plugins/drone-docker:<NEW_VERSION>
    repo: <MY_REPO>
    use_cache: true
    volumes:
```

Needs to be run as trusted for the volume obviously!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,3 @@ _testmain.go
 
 coverage.out
 drone-docker
-
-# IDE/Editor related files
-**.swp
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ _testmain.go
 
 coverage.out
 drone-docker
+
+# IDE/Editor related files
+**.swp
--- a/main.go
+++ b/main.go
@@ -118,6 +118,11 @@ func main() {
 			Usage:  "squash the layers at build time",
 			EnvVar: "PLUGIN_SQUASH",
 		},
+		cli.BoolTFlag{
+			Name:   "cache",
+			Usage:  "don't attempt to re-build layers of the image that already exist",
+			EnvVar: "PLUGIN_USE_CACHE",
+		},
 		cli.BoolFlag{
 			Name:   "compress",
 			Usage:  "compress the build context using gzip",
@@ -172,6 +177,7 @@ func run(c *cli.Context) error {
 			Tags:       c.StringSlice("tags"),
 			Args:       c.StringSlice("args"),
 			Squash:     c.Bool("squash"),
+			Cache:       c.Bool("cache"),
 			Compress:   c.Bool("compress"),
 			Repo:       c.String("repo"),
 		},
--- a/main.go
+++ b/main.go
@@ -119,9 +119,9 @@ func main() {
 			EnvVar: "PLUGIN_SQUASH",
 		},
 		cli.BoolTFlag{
-			Name:   "cache",
-			Usage:  "don't attempt to re-build layers of the image that already exist",
-			EnvVar: "PLUGIN_USE_CACHE",
+			Name:   "pull-image",
+			Usage:  "force pull base image at build time",
+			EnvVar: "PLUGIN_PULL_IMAGE",
 		},
 		cli.BoolFlag{
 			Name:   "compress",
@@ -177,7 +177,7 @@ func run(c *cli.Context) error {
 			Tags:       c.StringSlice("tags"),
 			Args:       c.StringSlice("args"),
 			Squash:     c.Bool("squash"),
-			Cache:       c.Bool("cache"),
+			Pull:       c.BoolT("pull-image"),
 			Compress:   c.Bool("compress"),
 			Repo:       c.String("repo"),
 		},
--- a/plugin.go
+++ b/plugin.go
@@ -47,7 +47,7 @@ type (
 		Tags       []string // Docker build tags
 		Args       []string // Docker build args
 		Squash     bool     // Docker build squash
-		Cache      bool     // Docker build without pulling
+		Pull       bool     // Docker build pull
 		Compress   bool     // Docker build compress
 		Repo       string   // Docker build repository
 	}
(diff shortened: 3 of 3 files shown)


plugin.go
```diff
@@ -200,6 +200,11 @@ func commandBuild(build Build) *exec.Cmd {
 	if build.Compress {
 		args = append(args, "--compress")
 	}
+    if build.Cache {
+        args = append(args, "--pull=false")
+    } else {
+        args = append(args, "--pull=true")
+    }
```
bradrydzewski: I think this could be condensed to a single line:

```
if build.Pull {
  args = append(args, "--pull=true")
}
```

plugin.go
```diff
@@ -47,6 +47,7 @@ type (
 		Tags       []string // Docker build tags
 		Args       []string // Docker build args
 		Squash     bool     // Docker build squash
+		Cache      bool     // Docker build without pulling
```
bradrydzewski: recommend change to `Pull`

main.go
```diff
@@ -172,6 +177,7 @@ func run(c *cli.Context) error {
 			Tags:       c.StringSlice("tags"),
 			Args:       c.StringSlice("args"),
 			Squash:     c.Bool("squash"),
+			Cache:       c.Bool("cache"),
```
bradrydzewski: recommend change to:

```
c.BoolT("pull-image")
```

main.go
```diff
@@ -118,6 +118,11 @@ func main() {
 			Usage:  "squash the layers at build time",
 			EnvVar: "PLUGIN_SQUASH",
 		},
+		cli.BoolTFlag{
+			Name:   "cache",
+			Usage:  "don't attempt to re-build layers of the image that already exist",
+			EnvVar: "PLUGIN_USE_CACHE",
```
bradrydzewski: recommend change to:

```
cli.BoolTFlag{
  Name: "pull_image",
  Usage: "pull a newer version of the image on build",
  EnvVar: "PLUGIN_PULL_IMAGE",
}
```

main.go
bradrydzewski: also note we need to use `c.BoolT` here

main.go
bradrydzewski: note that we cannot use `pull` here. This is a reserved keywork in the yaml and will not get passed to the plugin. So instead we use `pull_image`

.gitignore
```diff
@@ -26,3 +26,6 @@ _testmain.go
 
 coverage.out
 drone-docker
+
+# IDE/Editor related files
+**.swp
```
bradrydzewski: can you revert this file? You can add editor specific files to `.git/info/exclude`

plugin.go
bradrydzewski: also make sure to run `gofmt` on your changes. This will ensure the codebase conforms to the official Go style guidelines (in this case tags for indentation)

JeffDownie: Doooh.. now I feel silly! I knew there was a way of doing this without negating the option, thanks!

bradrydzewski:
I should also mention that mounting a volume should not be required, and may result in unexpected behavior and certain settings not working because you are bypassing the docker daemon that is executing inside this plugins container. 

This plugin is meant to give the ability to build and publish images without compromising the host machine. So if you are mounting a docker volume, I would recommend not using this plugin and instead running docker commands directly:

```
pipeline:
  publish:
-   image: plugins/docker
+   image: docker
    commands:
      - docker build .
      - docker tag ...
      - ...
    volumes:
```

If you want to use the docker plugin with layer caching, the recommended solution is to use one of the drone caching plugins. These plugins give you the ability to cache files and folders and restore them at the start of your build.

```yaml
pipeline:
  restore_cache:
    image: <cache plugin>
    restore: true
    mount: /drone/docker
  publish:
    image: plugins/docker
    repo: foo/bar
    storage_path: /drone/docker  # volume shared by all containers in the pipeline
  rebuild_cache:
    image: <cache plugin>
    restore: true
    mount: /drone/docker
```

In the above example we make sure the docker storage path is in the shared volume so that is accessible by all steps in the pipeline. The first step (restore) will restore the docker storage path from the cache. The last step will update it. This is the recommended way to persist state between builds without having to mount a volume (although there is a cache plugin that does this too, if you want a local cache)

JeffDownie:
@bradrydzewski I've made the changes - didn't know about this gofmt tool, nice!

It might be worth mentioning, I had a lot of trouble running my own version of this plugin on my own drone server, because it is [hardcoded to run as privileged in the actual drone agent](https://github.com/drone/drone/blob/master/drone/exec.go#L73) - it might be worth documenting this somewhere!

I also came across an issue where I tried to make the `storage_path` a relative path, which caused the docker daemon to hang on startup. (I wanted to store the docker stuff in a subdirectory of the workspace, so it would definitely be shared between everything in the pipeline)

JeffDownie:
@bradrydzewski I've been trying to use the above caching method with this plugin, and I can't seem to get it to work nicely - how can you avoid the cache growing in size with old images continually? There's no obvious way of cleaning the docker storage of everything _except_ an image?

The only obvious solution that comes to my mind is not using any drone caching plugin and instead pulling down a previous version of the image from the registry on this image's startup. This feels a little excessive for a plugin though - I may just write my own commands onto a docker-in-docker container to get the caching strategy I need.

bradrydzewski:
maybe add a `prune: true` flag to the yaml, and if true it runs `docker image prune`. This will remove all dangling layers. Some cache plugins also let you control the duration of the cache (ie 30 days). Pruning and cache expiration can both probably help reduce this problem.

> I may just write my own commands onto a docker-in-docker container to get the caching strategy I need.

That is fine too. This is not intended to be a one-size-fits-all plugin. I personally encourage people to fork or create alternate versions of plugins tailored to fit their specific requirements and use cases.

jamrizzi:
@bradrydzewski, what cache plugin should I use for `<cache plugin>`?

PR #114, opened 2017-03-10
Update docker to 17.03

tuxity:
There is no major changes, only bugfix. We will be up to date with new docker version system.

https://github.com/docker/docker/releases/tag/v17.03.0-ce

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,7 +16,7 @@ pipeline:
   publish:
     image: plugins/docker
     repo: plugins/docker
-    tags: [ "latest", "1.13", "1" ]
+    tags: [ "latest", "17.03", "17" ]
     when:
       branch: master
       event: push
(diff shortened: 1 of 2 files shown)


tuxity:
@donny-dont they said : 

```
Starting with this release, Docker is on a monthly release cycle and uses a
new YY.MM versioning scheme to reflect this. Two channels are available: monthly and quarterly.
Any given monthly release will only receive security and bugfixes until the next monthly
release is available. Quarterly releases receive security and bugfixes for 4 months after
initial release. 
```

bradrydzewski:
I'm fine with this, but we would need to implement the following https://github.com/drone-plugins/drone-docker/issues/116 and we would need to store the version in a `VERSION` file

tuxity:
Closing in favor of #122

PR #118, opened 2017-03-16
Change env precedence

greysond:
When using the config values `registry`, `username`, `password`, and `email`, if the corresponding `DOCKER_` env var is set, the explicit values are ignored. This is due to the `urfave/cli` package taking the first available env var.

This PR changes the order so that explicit configuration is favored over environmental. If the config order is intentional, please feel free to close this PR!

Thanks in advance!

### DIFF
--- a/main.go
+++ b/main.go
@@ -137,22 +137,22 @@ func main() {
 			Name:   "docker.registry",
 			Usage:  "docker registry",
 			Value:  defaultRegistry,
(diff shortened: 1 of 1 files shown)


jmccann:
This makes sense to me as interpolation is the future and injection is going away ... so give precedence to values provided via `.drone.yml` directly.

PR #119, opened 2017-03-22
Remove temp tag from build and prune afterwards

jmccann:
This PR will untag the "temp" tag (`p.Build.Name`) from the image so that `docker system prune` can properly clean it up.  Currently `docker system prune` does nothing.

However, this PR will not work until Docker 17.04 is released as there is a bug prior to that where `docker system prune` does not properly clean dangling images.  https://github.com/docker/docker/issues/32006

So I just wanted to get this code/issue in the open.  We can either merge knowing it won't really do anything until 17.04 is released and the plugin uses it ... or wait until 17.04 is released and the plugin is using it before merging.

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -112,7 +112,7 @@ func (p Plugin) Exec() error {
 	var cmds []*exec.Cmd
 	cmds = append(cmds, commandVersion())      // docker version
 	cmds = append(cmds, commandInfo())         // docker info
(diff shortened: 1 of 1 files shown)


bradrydzewski:
thanks, do you want to update the PR to include the changes to 17.04?

jmccann:
Now that #122 and #124 have been merged this should be ready to go.

PR #129, opened 2017-06-16
integrate changes from drone-gcr/pull/20

stephansnyt:
This is based on work here: https://github.com/drone-plugins/drone-gcr/pull/20

I made a couple changes to the wrapper script, most are pretty inconsequential but changing the way that jq is called to "detect" json in the token value was important: the previous version would leak the token into build logs regardless of whether debug was turned on.

I had to make a change to `plugin.go` to concatenate the docker registry and repo together. I assumed that if you did a `docker login` to `gcr.io` it would pick that registry for a `docker push` but that was not the case.

Also I'm not exactly sure how this is going to work w.r.t. tags. `Dockerfile_gcr` wants to pull `FROM plugins/docker:17.05` but that depends on the previous step running, and updating the image under that tag.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -21,3 +21,14 @@ pipeline:
     when:
       branch: master
       event: push
+
+
+  publish:
+    image: plugins/gcr
+    repo: plugins/gcr
+    dockerfile: Dockerfile_gcr
+    tags: [ "latest", "17", "17.05" ]
+    secrets: [ docker_username, docker_password ]
+    when:
+      branch: master
+      event: push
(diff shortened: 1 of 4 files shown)


bradrydzewski:
thanks! sorry for the delay. I ended up using this as a baseline and implementing in pure Go. The reason is that pure Go implementations end up making it easier for us to publish for multiple architectures (arm and arm64) which are supported as of 0.8

you can see the Go implementation of your bash script at
https://github.com/drone-plugins/drone-docker/blob/master/cmd/drone-docker-gcr/main.go

the image is now published to plugins/gcr

PR #135, opened 2017-07-17
Integrated label-schema.org labels

### LINKED ISSUE #134: Support optionally label-schema.org labels
The project http://label-schema.org/ is working on establishing standard Docker labels for all Docker images. It's a community-driven label namespace.

Services like https://microbadger.com/labelschema are already displaying these labels within the layer overview, take https://microbadger.com/images/owncloud/server as an example.

There are some labels we already know while we are building the Docker image like `build-date`, `vcs-url`, `vcs-ref` or `version`. Other labels can be defined like that as an example:

```diff
pipeline:
  docker:
    image: plugins/docker
+    label_schema:
+      name: Awesome Tool
+      description: This is the description
+      vendor: Thomas Boerger Inc.
```

A decision has to be made for the way to enable the labels. Do we want another attribute like `label_schema_defaults: true`? Or do we want to set all labels if the `label_schema` key is already present?

The example for Docker Hub uses build args, but since we are in the control of the docker build command we can directly set `--label` flags.

tboerger:
Fixes #134


tboerger:
## Command

```
docker run -ti --rm -v $(pwd):$(pwd) -w $(pwd) --privileged plugins/docker --dry-run --remote.url git@github.com:drone-plugins/drone-docker.git --label-schema name=Docker --repo plugins/docker
```

## Output

```
<SNIP>
+ /usr/local/bin/docker build --rm=true -f Dockerfile -t 00000000 . --pull=true --label org.label-schema.build-date=2017-07-17T19:49:21Z --label org.label-schema.vcs-ref=00000000 --label org.label-schema.vcs-url=git@github.com:drone-plugins/drone-docker.git --label org.label-schema.name=Docker

<SNIP>

 ---> b028262b982d
Removing intermediate container 00a95f6f4d2a
Step 6/6 : LABEL org.label-schema.build-date '2017-07-17T19:49:21Z' org.label-schema.name 'Docker' org.label-schema.vcs-ref '00000000' org.label-schema.vcs-url 'git@github.com:drone-plugins/drone-docker.git'
```

plugin.go
```diff
@@ -200,6 +202,20 @@ func commandBuild(build Build) *exec.Cmd {
 		args = append(args, "--build-arg", arg)
 	}
 
+	labelSchema := []string{
+		fmt.Sprintf("build-date=%s", time.Now().Format(time.RFC3339)),
+		fmt.Sprintf("vcs-ref=%s", build.Name),
+		fmt.Sprintf("vcs-url=%s", build.Remote),
+	}
+
```
bradrydzewski: shoule we also add `org.label-schema.schema-version = "1.0"`

> All labels are OPTIONAL, although org.label-schema.schema-version SHOULD be present.

plugin.go
```diff
@@ -200,6 +202,20 @@ func commandBuild(build Build) *exec.Cmd {
...
+	if len(build.LabelSchema) > 0 {
+		labelSchema = append(labelSchema, build.LabelSchema...)
+	}
+
+	for _, label := range labelSchema {
+		args = append(args, "--label", fmt.Sprintf("org.label-schema.%s", label))
```
bradrydzewski: should we automatically prefix the `org.label-schema` for the user?  This plugin generally forces explicit values instead of implicit (for example, the repo field is not automatically prefixed with the registry hostname). I don't have a strong opinion either way, so wanted to pose the question. If you think this is the best approach, then no changes required :)

plugin.go
tboerger: Yes, you are right, I have to add it.

plugin.go
tboerger: I'm open for ideas, we can also name the flag only labels and allow general label addition, not just labels prefixed with `org.label-schema.`. We can set the 4 labels from above automatically as we know the values for them, otheres have to be set like this in that case:

```
pipeline:
  docker:
    image: plugins/docker
    pull: true
    secrets: [ docker_username, docker_password ]
    repo: foo/foo
    labels:
      - org.label-schema.name=Foo
      - org.label-schema.version=${DRONE_BRANCH}
```

bradrydzewski:
I'm going to merge as-is. I think it is a solid implementation and I don't want to hold it back :)

PR #136, opened 2017-08-04
update docker to support multistage builds and add build target flag

sethpollack:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker:17.05.0-ce-dind
+FROM docker:17.07.0-ce-dind
```
bradrydzewski: when bumping the docker version, we should also bump the tag in the .drone.yml

```diff
  publish:
    image: plugins/docker
    repo: plugins/docker
-   tags: [ "latest", "17", "17.05" ]
+   tags: [ "latest", "17", "17.07" ]
```

sethpollack:
Closing this in favor of #161, I will open a separate PR for the build target flag

PR #142, opened 2017-08-28
Configurable pruning

gtaylor:
As per Gitter [convo](https://gitter.im/drone/drone?at=59a434ff210ac26920cdc5e9), this PR adds flags for disabling post-build rmi and pruning. We'll definitely want to do image removal out-of-band, but may also want to handle pruning on our own. This is a Kubernetes setup with a separate DinD container co-located in the pod that contains the agent, which means we're still DinD (but not within Drone).

cc @foklepoint

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -102,6 +102,11 @@ func main() {
 			Usage:  "don't start the docker daemon",
 			EnvVar: "PLUGIN_DAEMON_OFF",
 		},
+		cli.BoolFlag{
+			Name:   "daemon.disable-system-prune",
+			Usage:  "don't run post-build docker system prune",
+			EnvVar: "PLUGIN_DISABLE_SYSTEM_PRUNE",
+		},
 		cli.StringFlag{
 			Name:   "dockerfile",
 			Usage:  "build dockerfile",
@@ -214,6 +219,7 @@ func run(c *cli.Context) error {
 			StoragePath:   c.String("daemon.storage-path"),
 			Insecure:      c.Bool("daemon.insecure"),
 			Disabled:      c.Bool("daemon.off"),
+			DisablePrune:  c.Bool("daemon.disable-system-prune"),
 			IPv6:          c.Bool("daemon.ipv6"),
 			Debug:         c.Bool("daemon.debug"),
 			Bip:           c.String("daemon.bip"),
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -107,6 +107,11 @@ func main() {
 			Usage:  "don't run post-build docker system prune",
 			EnvVar: "PLUGIN_DISABLE_SYSTEM_PRUNE",
 		},
+		cli.BoolFlag{
+			Name:   "daemon.disable-image-removal",
+			Usage:  "don't run post-build docker rmi",
+			EnvVar: "PLUGIN_DISABLE_IMAGE_REMOVAL",
+		},
 		cli.StringFlag{
 			Name:   "dockerfile",
 			Usage:  "build dockerfile",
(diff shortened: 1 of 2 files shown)


cmd/drone-docker/main.go
```diff
@@ -102,6 +102,16 @@ func main() {
 			Usage:  "don't start the docker daemon",
 			EnvVar: "PLUGIN_DAEMON_OFF",
 		},
+		cli.BoolFlag{
```
tboerger: Maybe use `BoolTFlag` so that it defaults to `true` and drop the `DISABLE` naming?

docker.go
```diff
@@ -122,8 +124,12 @@ func (p Plugin) Exec() error {
 		}
 	}
 
-	cmds = append(cmds, commandRmi(p.Build.Name)) // docker rmi
-	cmds = append(cmds, commandPrune())           // docker system prune -f
+	if !p.Daemon.DisableRMI {
+		cmds = append(cmds, commandRmi(p.Build.Name)) // docker rmi
```
tboerger: I know it's not part of this PR, but what about using `docker image prune` instead of removing all?

cmd/drone-docker/main.go
gtaylor: I implemented this per the above linked Gitter convo with @bradrydzewski, so I'll defer to him on that. Happy to make the change if he's cool with it.

docker.go
gtaylor: I think that definitely be neat, but would rather that be another discussion and subsequent PR.

tboerger: Yeah sure, just wanted to mention that because I have not seen this snippet before :)

bradrydzewski:
I'm just trying to figure out if we want to disable `docker rmi`. Do you want this here to avoid rebuilding layers? If yes, we should run some tests to make sure it works as expected. Also if we disable `rmi` we probably at least want to remove dangling images to avoid buildup

gtaylor:
Since we'll be using an external-to-this-plugin docker daemon (but Pod-colocated), we'll handle rmi and purging separately from the plugin. My goal is for the plugin just to be concerned with building and pushing for our usage case.

bradrydzewski:
I just want to understand the reason for not removing the temporary build image

bradrydzewski:
perhaps instead lets just disable `system prune` and then if you decide you need to disable `rmi` you can send a second pull request? I would prefer only adding flags once we have a documented use case for why it is needed.

gtaylor:
We prefer to take advantage of the image layer cache for subsequent builds. We have some very large/slow builds that will benefit from this. We can also do a pod layout where we have one dind container and 3-4 agents, helping us improve our cache hit rate and lower our number of duplicated (across docker daemons) things.

This assumes that each agent isn't mucking with our docker daemon, though (aside from building stuff).

bradrydzewski:
OK cool, thanks for the clarification. I just have two minor requests, if you don't mind ...

First, can you test and confirm that layer caching works properly by running the plugin locally multiple times, with different commit hashes each time? So something like this:

* using the example at https://github.com/drone-plugins/drone-docker#usage
* set DRONE_COMMIT_SHA=...
* set PLUGIN_DAEMON_OFF=true
* mount the host machine docker socket

Second, if we disable `docker rmi` should we remove dangling images? I feel like we might want to. This shouldn't conflict with your own internal garbage collection (so to speak) but might allow others to use this feature without having to implement their own garbage collection.

tboerger:
docker image prune -f would help to prune it

codycraven:
It looks like this has been handled differently,

```go
		cli.BoolTFlag{
			Name:   "docker.purge",
			Usage:  "docker should cleanup images",
			EnvVar: "PLUGIN_PURGE",
		},
```

```go
	if p.Cleanup {
		cmds = append(cmds, commandRmi(p.Build.Name)) // docker rmi
		cmds = append(cmds, commandPrune())           // docker system prune -f
	}
```

Is there a reason to keep this PR open?

bradrydzewski:
@gtaylor sorry, looks like this ended up getting merged as part of another PR. You can implement with `purge: true`

PR #144, opened 2017-08-29
Change possible password fields for GCR

tonglil:
GOOGLE_APPLICATION_CREDENTIALS "should point to a file that defines the credentials", and should not be the contents of the credentials to follow Google's convention.

See https://developers.google.com/identity/protocols/application-default-credentials#howtheywork.

I replaced it with `TOKEN` to match the parameter of the previous GCR plugin worked, which may help with migrations.

### DIFF
--- a/cmd/drone-docker-gcr/main.go
+++ b/cmd/drone-docker-gcr/main.go
@@ -19,7 +19,7 @@ func main() {
 			"PLUGIN_JSON_KEY",
 			"GCR_JSON_KEY",
 			"GOOGLE_CREDENTIALS",
-			"GOOGLE_APPLICATION_CREDENTIALS",
+			"TOKEN",
 		)
 	)
 


cmd/drone-docker-gcr/main.go
```diff
@@ -19,7 +19,7 @@ func main() {
 			"PLUGIN_JSON_KEY",
 			"GCR_JSON_KEY",
 			"GOOGLE_CREDENTIALS",
-			"GOOGLE_APPLICATION_CREDENTIALS",
+			"TOKEN",
```
tboerger: But we are always prefixing this kind of env variable names.

tonglil: Do you mean to prefix it with `PLUGIN_TOKEN`?

I was thinking about this from a `.drone.yml` standpoint.

0.4:
```yaml
publish:
  gcr:
    repo: foo/bar
    token: >
      $$TOKEN
```

0.7+:
```yaml
pipeline:
  gcr:
    image: plugins/gcr
    repo: foo/bar
    secrets: [token]
```

But I'm ok with removing `TOKEN` as well :)

bradrydzewski:
I'm fine with keeping for backward compatibility in the short term. Lets plan on removing token in the coming months prior to a 1.0 release, and we can add to the release notes to give everyone a proper heads up.

PR #145, opened 2017-09-04
Drone ECR plugin

donkeysharp:
Publish docker images to an ECR registry


cmd/drone-docker-ecr/main.go
```diff
@@ -1,5 +1,113 @@
...
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"os"
+	"os/exec"
+)
+
+const (
+	DOCKER_USER    = "AWS"
```
tboerger: Constants should be written in camel case like DoxkerUser

cmd/drone-docker-ecr/main.go
```diff
@@ -1,5 +1,113 @@
...
+	DEFAULT_REGION = "us-east-1"
```
tboerger: Same here

cmd/drone-docker-ecr/main.go
```diff
@@ -1,5 +1,113 @@
...
 func main() {
+	var registryIds []*string
+
+	ecrRegion := DEFAULT_REGION
+
+	if getenv("ECR_REGION") != "" {
+		os.Setenv("PLUGIN_REGION", getenv("ECR_REGION"))
+	}
+
+	if accessKey := getenv("ECR_ACCESS_KEY"); accessKey != "" {
```
tboerger: You should call getenv with all possible env names, not one by one, look at the gcr command

docker/ecr/Dockerfile
```diff
@@ -1 +1,6 @@
 FROM plugins/docker:latest
+
+RUN apk update && apk add ca-certificates
```
tboerger: This should be already done for plugins/docker so you can skip that

donkeysharp: I thought that was already installedbut it isn't.
Is it ok if I move `ca-certificates` to `docker/Dockerfile`? so other plugins have that package as well

tboerger: IMHO it have to be part of the base image, so it's plugins/docker

donkeysharp:
Update changes according to review

cmd/drone-docker-ecr/main.go
```diff
@@ -11,46 +11,40 @@ import (
...
-		os.Setenv("PLUGIN_ACCESS_KEY", accessKey)
-	}
-
-	if secretKey := getenv("ECR_SECRET_KEY"); secretKey != "" {
-		os.Setenv("PLUGIN_SECRET_KEY", secretKey)
-	}
-
-	if accessKey := getenv("PLUGIN_ACCESS_KEY"); accessKey != "" {
+	if accessKey != "" && secretKey != "" {
 		os.Setenv("AWS_ACCESS_KEY_ID", accessKey)
```
tboerger: This should also be moved to the top getenv call

cmd/drone-docker-ecr/main.go
```diff
@@ -1,5 +1,107 @@
...
+		secretKey  = getenv("ECR_SECRET_KEY", "PLUGIN_SECRET_KEY")
+		registries = getenv("ECR_REGISTRY_IDS", "PLUGIN_REGISTRY_IDS")
+	)
+
+	if ecrRegion == "" {
+		ecrRegion = DefaultRegion
+	}
+
+	if accessKey != "" && secretKey != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", accessKey)
```
tboerger: This should also be moved to the top getenv call

cmd/drone-docker-ecr/main.go
donkeysharp: I don't fully understand, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are read by AWS SDK http://docs.aws.amazon.com/sdk-for-go/api/aws/session/#hdr-Environment_Variables

So if we expect them to only come as environment variables e.g. as secrets, then this code has to be removed.
```		
accessKey  = getenv("ECR_ACCESS_KEY", "PLUGIN_ACCESS_KEY")
secretKey  = getenv("ECR_SECRET_KEY", "PLUGIN_SECRET_KEY")
```

If that the case, it must be documented so user knows how to call the secrets i.e. `aws_access_key_id` and `aws_secret_access_key`. In addition it won't be possible to use `PLUGIN_ACCESS_KEY` and `PLUGIN_SECRET_KEY`

donkeysharp: Anyhow I will read if it's possible to pass both access and secret keys as strings programatically, which would solve the problem.

tboerger: Sorry, was reading it wrong, I think setting the env variables should be fine

donkeysharp: so is it ok as it's now? or would it be better using credentials? - which would be something like: https://gist.github.com/donkeysharp/1949f93e44011c5df83ad4112879293b (I didn't tested yet)

tboerger:
@bradrydzewski any comments for that?

donkeysharp:
I updated to [use static credentials](https://github.com/drone-plugins/drone-docker/pull/145/commits/bba236f310958d83f9fd78def824ac8535d11acf), so there is no need to set environment variables in Go code. From my point of view it's a cleaner approach

ozbillwang:
@donkeysharp & @tboerger 

I am not Golang Developer, but it doesn't make sense for me to save other repositories in this repo under vendor folder. So it makes this commit with so many files to be changed. 

Are there any locks (such as Gemfiles, requirements.txt) or `git submodules` we can use to reduce this type of changes?

tboerger:
> I am not Golang Developer, but it doesn't make sense for me to save other repositories in this repo under vendor folder. So it makes this commit with so many files to be changed.

This is pretty conventional within Golang. There are some kind of lock files, but we already had some cases in the past where the source repo have been deleted and we have not been able to build the plugin anymore. All plugins and also Drone itself stores the vendoring within the repo. The used vendoring tool (govendor) already reduces the source files as it just keeps the files that are really imported and used.

And git submodules? Who is still using this crappy solution these days?

tboerger:
It's up to @bradrydzewski

bradrydzewski:
Until we can cryptographically verify large vendor updates (coming soon to dep) this is something I will need to manually download and merge. I plan on doing this soon, but haven't found the time yet. I will post an update as soon as available.

bradrydzewski:
I pulled down changes locally and pushed e4bc243c9350d2ee878be4d8046888628d7998eb. I switched to dep so that in the future we will not have to worry about vendored files and verifying large pull requests.

PR #148, opened 2017-09-12
Added implementation for Docker in Hyper containers

LFlare:
This will allow for building docker images on hyper containers. There has to be a better way of implementing this with `plugins/docker` but I have no idea what. Would appreciate feedback from you @bradrydzewski.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,6 +17,16 @@ pipeline:
       branch: master
       event: push
 
+  publish_hyper:
+    image: plugins/docker
+    repo: plugins/hyper
+    tags: "latest"
+    secrets: [ docker_username, docker_password ]
+    dockerfile: docker/hyper/Dockerfile
+    when:
+      branch: master
+      event: push
+
   publish_heroku:
     image: plugins/docker
     repo: plugins/heroku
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -17,6 +17,16 @@ pipeline:
       branch: master
       event: push
 
+  publish:
```
tboerger: Wrong step name

.drone.yml
```diff
@@ -17,6 +17,16 @@ pipeline:
...
+    image: plugins/docker
+    repo: plugins/docker
```
tboerger: I don't think that brad want's to add this to the same image as this is generally feature-frozen.

LFlare:
@tboerger, thanks for the feedback, I made some typos and after some considerations, I think `plugins/hyper` is a better choice, what do you think?

bradrydzewski:
I recommend starting by creating your own `LFlare/hyper` image. The process we generally following is that all plugins __must__ incubate as third party plugins before we would consider adding the the official plugins organization.

LFlare:
Alright, I've done it through `LFlare/drone-hyper` for now.

PR #149, opened 2017-09-12
Add support for --cache-from directive

benglewis:


bradrydzewski:
merged via https://github.com/drone-plugins/drone-docker/pull/206

PR #151, opened 2017-09-26
add cache-from

sethpollack:


docker.go
```diff
@@ -100,6 +101,15 @@ func (p Plugin) Exec() error {
 		fmt.Println("Registry credentials not provided. Guest mode enabled.")
 	}
 
+	// pre-pull cache image
+	if p.Build.CacheFrom != "" {
```
tuxity: Why pre-pulling it ? you are going to slow the plugin no ?

tuxity: oh no nvm

tuxity: But I'm not sure if it's useful with --pull param

sethpollack: `--cache-from` gets ignored if the image is not local

sethpollack: It's actually an array, I'll push an update.

cmd/drone-docker/main.go
```diff
@@ -131,6 +131,11 @@ func main() {
 			Usage:  "build args",
 			EnvVar: "PLUGIN_BUILD_ARGS_FROM_ENV",
 		},
+		cli.StringSliceFlag{
+			Name:   "cache-from",
+			Usage:  "cache from",
```
tuxity: `images to consider as cache sources` would be better :P https://docs.docker.com/engine/reference/commandline/build/

tuxity:
DDo you have an already built image in a registry to test it ?

sethpollack:
I have ones that includes all my open PR's 
`beenverifiedinc/drone_docker`
`beenverifiedinc/drone_ecr`

tuxity:
Okay thank, gonna test that in few minutes :)

tuxity:
ping @bradrydzewski @tboerger

tboerger:
Posts like https://medium.com/@gajus/making-docker-in-docker-builds-x2-faster-using-docker-cache-from-option-c01febd8ef84 do not mention any docker pull, is this really required?

Jean85:
Yep, pulling the image (or having the layers of those images already available) is required. No pull is triggered by Docker in the `--cache-from` subject is missing.

docker.go
```diff
@@ -100,6 +101,15 @@ func (p Plugin) Exec() error {
 		fmt.Println("Registry credentials not provided. Guest mode enabled.")
 	}
 
+	// pre-pull cache image
+	for _, img := range p.Build.CacheFrom {
+		cmd := commandPull(img)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		trace(cmd)
+		cmd.Run()
```
bradrydzewski: I think these commands should be appended to the list, after docker info

```diff
cmds = append(cmds, commandInfo())    // docker info

+for _, img := range p.Build.CacheFrom {
+  cmds = append(cmds, commandPull(img)) 
+}
```

bradrydzewski: I think these commands should be appended to the list, after docker info

```diff
cmds = append(cmds, commandInfo())    // docker info

+for _, img := range p.Build.CacheFrom {
+  cmds = append(cmds, commandPull(img)) 
+}
```

sethpollack: The problem with that is that it will throw an error when the image/tag doesn't exist yet.

Jean85: Maybe we can append them with `|| true` to avoid build failures? I do that manually in my CI for this specific reason.

bradrydzewski: @Jean85 I appreciate the suggest, however, `||` would not work because that is a shell notation and these commands are not executed with bash. They are executed from Go with the `exec` syscall.

sethpollack: @bradrydzewski do you still want me to move the commands?

tuxity: @bradrydzewski ping

tuxity:
bump @sethpollack @bradrydzewski It's a useful feature :(

docker.go
tuxity: I think you should move that command yes, the user have to enable the feature. If the name is wrong, the build should fail and the user should correct it

sethpollack: It's not a naming issue, its when you build the first time the image does not exist yet.

Sharsie:
@bradrydzewski Without seeing the pull request we have made a similiar conclusion as @sethpollack about pulling the images: The pull command needs to pass even with errors

Our reasoning was that you can really use variable cache_from value based on environment variables, such as `DRONE_COMMIT_BRANCH`, `DRONE_COMMIT_AUTHOR` or other similiar logic

In that case you definitely do not want the build to fail and want to have it create the tag for the first time, then use the image from previous builds as a cache source. This can include release and other branches. 

If you wish to test it using an already built plugin, you are welcome to our version: `yoursystemcz/drone-docker`

It is almost the same as this PR and we would welcome if it was merged as soon as possible so we can use the official docker plugin instead.


There would be an issue when using variables, such as push requests to hotfix branches, e.g. hotfix/Test, which would result in an image name "image:hotfix/Test" which violates tag names. 

This could be solved using a simple regex replace with dashes, I am willing to submit a PR for this if it is going to be merged.

Also, if you insist on putting the pull command after `info` and `version`, we would need to remove these from the cmds array and run them before pull manually. If this is something you would prefer, I can submit a PR as well.

sheerun:
cheering for finishing this pr

blackjid:
hey @bradrydzewski love this PR!, please consider it

docker.go
```diff
@@ -195,6 +209,9 @@ func commandBuild(build Build) *exec.Cmd {
 	if build.Pull {
 		args = append(args, "--pull=true")
 	}
+	for _, arg := range build.CacheFrom {
+		args = append(args, "--cache-from", arg)
```
marcbachmann: Does this still succeed in case all the declared images are missing?
I had a similar issue in my current travis setup with bash and used `${CACHED_IMAGE:-false}` as fallback argument to get it to work:
https://gist.github.com/marcbachmann/16574ba8c614bb3b78614a351f324b86#file-travis-docker-scripts-md

tboerger:
Right, forgot about that. Thanks for the preparation of this feature.

PR #152, opened 2017-09-26
add ecr support

sethpollack:


sethpollack:
Already implemented in #155

PR #157, opened 2017-10-25
use dockerd data-root option

danielkrainas:
The daemon's `-g` and `--graph` options are deprecated as of 17.05, `--data-root` is the recommended alternative going forward. See moby/moby#28696


jpds:
This is still valid, can we please merge this in? @bradrydzewski

PR #159, opened 2017-11-01
Allow ecr login with default aws credential chain

AKoetsier:
Currently you have to specify static AWS credentials, this patch allow the sdk to use the instance- or container-role again.

### DIFF
--- a/cmd/drone-docker-ecr/main.go
+++ b/cmd/drone-docker-ecr/main.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ecr"
 )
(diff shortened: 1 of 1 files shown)


cmd/drone-docker-ecr/main.go
```diff
@@ -48,6 +50,7 @@ func main() {
 	svc := ecr.New(sess)
 	username, password, registry, err := getAuthInfo(svc)
 	if err != nil {
+		log.Fatal(fmt.Sprintf("error getting ECR auth: %v", err))
```
gtaylor: If we're using log.Fatal here, no need for the following os.Exit(1) anymore.

cmd/drone-docker-ecr/main.go
```diff
@@ -32,14 +32,16 @@ func main() {
...
-	os.Setenv("AWS_SECRET_ACCESS_KEY", secret)
 	os.Setenv("AWS_REGION", region)
+	awsConfig := aws.Config{Region: &region}
 
-	sess, err := session.NewSession(&aws.Config{
-		Credentials: credentials.NewEnvCredentials(),
-		Region:      &region,
-	})
+	if key != "" && secret != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", key)
```
gtaylor: I had thought that the AWS SDK had built-in understanding of these environment variables if you use the default credential chain? Are we sure that we even need to be handling them at all?

AKoetsier: I removed the extra credential provider. You are right the default chain will use them when available.

AKoetsier: You're right, I removed the explicit credential provider and allow the default chain to find the correct credentials.

Issue #150, opened 2017-09-15
Implement default tags logic


bradrydzewski:
I would like to detail my plan to implement an attribute called `default_tags`

```
docker:
  image: plugins/docker
  default_tags: true
```

When this feature is enabled and the event type is tag, drone will automatically tag the image using the standard major, minor, release convention. Some examples:

* `1.0.0` produces docker tags `1`, `1.0`, `1.0.0`
* `1.0.0-rc.1` produces docker tags `1.0.0-rc.1`

If default tags are enabled but the tag is not valid semver ad cannot be parsed, the plugin will default to the `latest` tag like it does today. This may be changed in the future and result in an error and the plugin exiting.

The primary benefit here is that right now we have a somewhat awkward workflow where you have to remember to update the tags in your yaml configuration file before tagging a release (assuming you want to follow this convention of tagging major, minor, release).

bradrydzewski:
Note I am open to other names ... `default_tag`, `auto_tag`, etc ...

tboerger:
Sounds good, but does it also work with v prefixes? Do they get dropped?

tonglil:
Should `-rc` releases be tagged to update the major, minor, or patch versions?

Also I assume `1.1` will only produce docker tags `1` and `1.1`?

ozbillwang:
not really agree. 

I hope at leave one tag need be unique. So every time, with a new build, new tag, for example, `1.0.<drone_build_number>` will be created and push to hub.docker.io

With that, I can lock my plugin in some version, if I don't want to use `latest` as tag.

bradrydzewski:
@tonglil 

> Should -rc releases be tagged to update the major, minor, or patch versions?

pre-releases will not create major, minor or patch versions. For example if we have `1.0.0-alpha.1` it will not create `1.0.0`, `1.0` and `1` tags. It will only create a `1.0.0-alpha.1` tag.

> Also I assume 1.1 will only produce docker tags 1 and 1.1?

`1.1` would create the default tags `1.0.0`, `1.0` and `1`

bradrydzewski:
@tboerger I am also adding a `default_suffix` parameter that can be used in conjunction with the `default_tags` parameter to help with multi-architecture images in the short term.

```
repo: golang
default_tags: true
default_suffix: nanoserver
```
 
[...]

PR #162, opened 2017-11-23
ignore empty tags to make plugin more stable

wavezhang:
There is a common use for this plugin to publish ```latest``` as well as ```${DRONETAG}```

```
    tags:
      - latest
      - ${DRONE_TAG}
```
However, sometimes code change without adding a tag will publish fail since there is no tags, but we need to publish to the ```latest``` in this case.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -116,6 +116,10 @@ func (p Plugin) Exec() error {
 	cmds = append(cmds, commandBuild(p.Build)) // docker build
 
 	for _, tag := range p.Build.Tags {
+		if tag == "" { // ignore empty tags
+			continue
+		}
+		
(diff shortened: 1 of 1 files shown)


bradrydzewski:
I think a blank tag should probably result in an error.

Note that you can use the emulated bash substitution functions to fallback to latest:

```diff
-    tags:
-      - latest
-      - ${DRONE_TAG}
+    tags:
+      - ${DRONE_TAG=latest}
```

Or you can use the new `default_tags: true` option which you can learn more about [here](https://github.com/drone-plugins/drone-docker/issues/150)

wavezhang:
That‘s a good trick. I'll have a try.

PR #166, opened 2018-01-23
Use stable v17.12 Docker dind image.

bweston92:
Uses the latest stable dind by default, which at the minute has better default ca certificates.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -13,7 +13,7 @@ pipeline:
   publish:
     image: plugins/docker
     repo: plugins/docker
-    tags: [ "latest", "17", "17.10", "17.12" ]
+    tags: [ "latest", "17", "17.12" ]
     secrets: [ docker_username, docker_password ]
     dockerfile: docker/Dockerfile
     when:
@@ -23,7 +23,7 @@ pipeline:
   publish_heroku:
     image: plugins/docker
     repo: plugins/heroku
-    tags: [ "latest", "17", "17.10", "17.12" ]
+    tags: [ "latest", "17", "17.12" ]
     secrets: [ docker_username, docker_password ]
     dockerfile: docker/heroku/Dockerfile
     when:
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # docker build --rm -f docker/Dockerfile -t plugins/docker .
 
-FROM docker:17.10.0-ce-dind
+FROM docker:stable-dind
```
bradrydzewski: We pin the plugin to a specific docker version and tag the plugin using the same version number as Docker. So we would accept `17.12.0-ce-dind` which is the same as `stable-dind` with a version bump in the `.drone.yml` for `17.12`

bweston92: @bradrydzewski thanks for the quick response, updated the PR.

.drone.yml
```diff
@@ -23,7 +23,7 @@ pipeline:
   publish_heroku:
     image: plugins/docker
     repo: plugins/heroku
-    tags: [ "latest", "17", "17.10" ]
+    tags: [ "latest", "17", "17.10", "17.12" ]
```
bradrydzewski: remove 17.10

```diff
-tags: [ "latest", "17", "17.10", "17.12" ]
+tags: [ "latest", "17", "17.12" ]
```

.drone.yml
```diff
@@ -13,7 +13,7 @@ pipeline:
   publish:
     image: plugins/docker
     repo: plugins/docker
-    tags: [ "latest", "17", "17.10" ]
+    tags: [ "latest", "17", "17.10", "17.12" ]
```
bradrydzewski: remove 17.10

```diff
-tags: [ "latest", "17", "17.10", "17.12" ]
+tags: [ "latest", "17", "17.12" ]
```

bradrydzewski:
thanks for the changes. 1 minor fix in the yaml, and I'll get it merged right away.

PR #161, opened 2017-11-20
Change base image from 17.05.0-ce-dind to stable-dind

delfer:
17.05.0 is not stable version and latest stable version 17.09 have some significant changes.

PR #168, opened 2018-02-07
Allow build for a specific target

mavimo:
In order to use the multi-stage build we should specify the `--target` flag (official [docs](https://docs.docker.com/engine/reference/commandline/build/#specifying-target-build-stage-target)), this PR allow us to indicate the layer to build, eg:
```diff
 pipeline:
   publish-dev:
     image: plugins/docker
     repo: mavimo/test
     tags: [ "latest", "prod" ]
+    target: production
     when:
       branch: master
       event: push

   publish-dev:
     image: plugins/docker
     repo: mavimo/test
     tags: [ "latest", "dev" ]
+    target: development
     when:
       branch: develop
       event: push
```
`Dockerfile`:
```
FROM php:7.1.13-fpm as production

COPY config/php.ini /usr/local/etc/php/php.ini

WORKDIR /srv

CMD ["php-fpm"]

FROM production as development

RUN apt-get update && apt-get install -y wget zip unzip git
```

NB: porting from https://github.com/drone-plugins/drone-docker/pull/136 and https://github.com/drone-plugins/drone-docker/pull/161

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -149,7 +149,7 @@ func main() {
 		cli.StringFlag{
 			Name:   "target",
 			Usage:  "build target",
-			EnvVar: "PLUGIN_BUILD_TARGET",
+			EnvVar: "PLUGIN_TARGET",
 		},
 		cli.BoolFlag{
 			Name:   "squash",
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -146,6 +146,11 @@ func main() {
 			Usage:  "build args",
 			EnvVar: "PLUGIN_BUILD_ARGS_FROM_ENV",
 		},
+		cli.StringFlag{
+			Name:   "target",
+			Usage:  "build target",
+			EnvVar: "PLUGIN_BUILD_TARGET",
+		},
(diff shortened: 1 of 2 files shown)


mavimo:
@bradrydzewski @tboerger do you think this should be useful? Any feedback?

mavimo:
@tboerger thx for the review, @bradrydzewski any concern from your side?

bradrydzewski:
LGTM can you also send a PR to https://github.com/drone/drone-plugin-index to document the new field? Thanks!

bradrydzewski:
Note that because the environment variable is PLUGIN_BUILD_TARGET you will need to use the `build_target` field in the yaml instead of `target`. If you would like to use `target` as the attribute name in the yaml, you will need to use PLUGIN_TARGET as the environment variable name.

```diff
 pipeline:
   publish-dev:
     image: plugins/docker
     repo: mavimo/test
     tags: [ "latest", "prod" ]
-    target: production
+    build_target: production
```

mavimo:
@bradrydzewski since other env variables mirroring the docker build flag I prefer to use `DRONE_TARGET` env variable and `target` flag. I'll open PR on docs if changes here are fine for you.

PR #169, opened 2018-02-10
Fix ECR registry string trimming

colinhoglund:
This addresses an issue when using `plugins/ecr` with `create_repository: true`. The current `TrimPrefix()` call tries to create a repo with a name like `/namespace/repo` which fails AWS API's [repositoryName constraints](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_CreateRepository.html#ECR-CreateRepository-request-repositoryName).

I believe this might address some of the issues @mhumeSF mentioned in [here](https://discourse.drone.io/t/plugins-ecr-does-not-auto-create-repo-in-ecr/1099).

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,8 @@ pipeline:
     commands:
       - go get -u github.com/golang/dep/cmd/dep
       - dep ensure
-      - go test ./...
+      - go vet ./...
+      - go test -cover ./...
       - sh .drone.sh
 
   publish:
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,6 +8,7 @@ pipeline:
     commands:
       - go get -u github.com/golang/dep/cmd/dep
       - dep ensure
+      - go test ./...
       - sh .drone.sh
 
   publish:
--- a/cmd/drone-docker-ecr/main.go
+++ b/cmd/drone-docker-ecr/main.go
@@ -55,7 +55,7 @@ func main() {
 	}
 
 	if create {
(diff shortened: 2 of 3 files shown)


cmd/drone-docker-ecr/main.go
```diff
@@ -75,6 +75,10 @@ func main() {
 	}
 }
 
+func getRepoName(repo string) string {
+	return strings.SplitN(repo, "/", 2)[1]
```
bradrydzewski: this could cause an index out-of-bounds panic

colinhoglund: Thanks for catching this.

.drone.yml
```diff
@@ -8,6 +8,7 @@ pipeline:
     commands:
       - go get -u github.com/golang/dep/cmd/dep
       - dep ensure
+      - go test ./...
```
tboerger: For other plugins we are using something like that to avoid testing of vendor directory:

```
  test:
    image: golang:1.9
    pull: true
    commands:
      - go vet
      - |
        for PKG in $(go list ./... | grep -v /vendor/); do
          go test -cover -coverprofile $GOPATH/src/$PKG/coverage.out $PKG
        done
```

bradrydzewski: @tboerger as of go 1.9 they ignore the vendor directory for tests :)

colinhoglund: I went ahead and pushed a commit to align build stage testing with other plugins.

mhumeSF:
This is awesome. A big hearty **thank you** for this!  ❤️ 👏 🎉

PR #174, opened 2018-03-07
Create image tags at build time

tuxity:
Instead of doing a `docker tag` at the end of the build, we can do it directly in the `docker build` command (since version 1.10)

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -124,10 +124,8 @@ func (p Plugin) Exec() error {
 
 	cmds = append(cmds, commandBuild(p.Build)) // docker build
 
-	for _, tag := range p.Build.Tags {
-		cmds = append(cmds, commandTag(p.Build, tag)) // docker tag
-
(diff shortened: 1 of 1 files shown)


tboerger:
Since 1.10 you can append multiple `-t` flags?

tuxity:
Yes, directly into docker build command.

https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t

Beanow:
Note, there's still a `docker rmi {build.Name}` command being executed at https://github.com/drone-plugins/drone-docker/pull/174/files#diff-d3a679229fa5a654b0e1d154531ac1e1R127

This change no longer builds an image with this reference name anymore though so it may not have the intended effect.

bradrydzewski:
before merging this, we should tag a new release so that we have something to rollback to if we encounter any unforseen issues.

PR #176, opened 2018-05-15
Bump docker/dind version for better CentOS support

RobinKnipe:
Fix errors encountered during `yum` install/update/clean etc.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 # docker build --rm -f docker/Dockerfile -t plugins/docker .
 
-FROM docker:17.12.0-ce-dind
+FROM docker:18.05.0-ce-dind
 
 ADD release/linux/amd64/drone-docker /bin/
 ENTRYPOINT ["/usr/local/bin/dockerd-entrypoint.sh", "/bin/drone-docker"]


mhumeSF:
@RobinKnipe Looking at the previous dind bump, I think tags have to be updated. https://github.com/drone-plugins/drone-docker/pull/161/files

tboerger:
@mrueg this plugin got tags for multiple different docker versions, just pin your builds to the version you want to have. If you are using no tag it's possible that your builds break because we are following more or less the latest versions.

docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # docker build --rm -f docker/Dockerfile -t plugins/docker .
 
-FROM docker:17.12.0-ce-dind
+FROM docker:18.05.0-ce-dind
```
bradrydzewski: also need to update the `.drone.yml` and bump all versions`18.05`

```diff
-tags: [ "latest", "17", "17.12" ]
+tags: [ "latest", "18", "18.05" ]
```

jpds:
I have a pull request with the requested changes at #194.

tboerger:
We are already publishing 18.06

PR #178, opened 2018-05-17
feat: handle race-condition under extreme use scenarios

zeyuye-airwallex:
# Overview
thanks for making such a handy plugin, we've come across an extreme scenerios when using `plugin/docker` and would like to share our changes to the main repo, we believe this would be helpful when to comes to optimising drone build speed while maintaining good reliability

## Summary on the changes

* use `satori (uuid.uuidv4())` for `-t` (`docker.Build.Name`)
* split `docker.purge` to `docker.purge.prune` and `docker.purge.named-tag` (the original `docker.purge` is still kept and can be used (as the global option) to enable all available cleanup types (`system prune` and `rmi named_tag`))

## Race Condition Scenarios
the extreme scenarios happens when the below two condition are satisfied
* parallel build job is enabled and multiple docker build using `plugin/docker` triggers at the same time in parallel
* host level `docker.sock` is use to docker build using `plugin/docker` (`/var/run/docker.sock` is mounted and `daemon_off` is true at the docker build step `plugin/docker`).

below is an example `.drone.yml` of the race condition scenario
```.yml
pipeline:
...
  # Build A
  build-a:
    group: docker-build
    docker:
    image: plugin/docker
    tags:
      - ${DRONE_REPO_BRANCH//\//-}-${DRONE_BUILD_NUMBER}
    repo: a
    dockerfile: a/Dockerfile
    context: ./a
    purge: false
    purge_named_tag: true
    daemon_off: true
    volumes:

  # Build B
  build-b:
    group: docker-build
    docker:
    image: plugin/docker
    tags:
      - ${DRONE_REPO_BRANCH//\//-}-${DRONE_BUILD_NUMBER}
    repo: b
    dockerfile: b/Dockerfile
    context: ./b
    purge: false
    purge_named_tag: true
    daemon_off: true
    volumes:
```

## Race Condition Explaination
When `plugin/docker` builds docker image

1. it uses `DRONE_COMMIT_SHA` as the `name:tag` to build the docker image first
2. it then retag the image (`DRONE_COMMIT_SHA`) to be `<repo>:<tags>`
3. it then push the image

this is fine if the `plugin/docker` drone jobs runs in its own isolated environment or runs sequentially (not in parallel)

however if `plugin/docker` drone jobs runs in parallel and shares the same `/var/run/docker.sock`, race condition occurs at the `tag` (step 2).

because all `plugin/docker` drone jobs builds the image using the same `name:tag` (`DRONE_COMMIT_SHA`), the `job2` (that finishes `docker build` step 1 later)'s `name:tag` could override the `job1` (that finishes `docker build` step 1 before)'s `name:tag` before the `job1` is able to complete its `tag` step 2. this means when `job1` does the tag, its actually tagging the image build by `job2`, hence this will cause the image that's pushed to registry of job1 is actually job2's image

by introducing the `name:tag` as `uuidV4()`, it avoids the `name:tag` collision (an impractical small possibility of collision), in addition, user can also specify the `named_tag` option to definite avoid the collision

```yml
pipeline:
  ...
  # Build B
  build-a:
    ...
    named_tag: a-0001121b17935768bb4a49dbc4d25821

  # Build B
  build-b:
    ...
    named_tag: b-0001121b17935768bb4a49dbc4d25821
```

## References
below shows the when each time `drone-docker` binary is initialised, it generates a random uuid for the `named.tag`

```bash
➜  drone-docker git:(develop) docker run -it --rm golang bash
root@d01cd4b90e60:/go# go get -u github.com/airwallex/drone-docker/cmd/drone-docker
root@d01cd4b90e60:/go# drone-docker --help
NAME:
   docker plugin - docker plugin
...
   --named.tag value              build name:tag (default: "1fcbcac9-75f1-4f7e-a3ee-d35a15ed2a68")
root@d01cd4b90e60:/go# drone-docker --help
NAME:
   docker plugin - docker plugin
...
   --named.tag value              build name:tag (default: "c40e65d6-f0b8-402c-bfbd-29ebed311238") 
root@d01cd4b90e60:/go# drone-docker --help
NAME:
   docker plugin - docker plugin
...
   --named.tag value              build name:tag (default: "1c21a81b-57f5-4c02-b5f1-00bfebfcdeaa") 
```

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -214,7 +214,7 @@ func main() {
 			Usage:  "docker should cleanup images",
 			EnvVar: "PLUGIN_PURGE_PRUNE",
 		},
-		cli.BoolFlag{
+		cli.BoolTFlag{
 			Name:   "docker.purge.named-tag",
 			Usage:  "docker should cleanup named tag",
 			EnvVar: "PLUGIN_PURGE_NAMED_TAG",
@@ -241,7 +241,7 @@ func run(c *cli.Context) error {
 		Dryrun:  c.Bool("dry-run"),
 		Cleanup: docker.Cleanup {
 			Prune:    c.BoolT("docker.purge") || c.Bool("docker.purge.prune"),
-			NamedTag: c.BoolT("docker.purge") || c.Bool("docker.purge.named-tag"),
+			NamedTag: c.BoolT("docker.purge") || c.BoolT("docker.purge.named-tag"),
 		},
 		Login: docker.Login{
 			Registry: c.String("docker.registry"),


tboerger:
If you change docker.go you must update the other commands as well.

Beside that your use case is not really the focus of that plugin. Not sure how @bradrydzewski things about that change.

zeyuye-airwallex:
thanks for looking into this @tboerger 

when you said `the other command`, what `command` are you refering to here.

tboerger:
Ah scratch that, the other plugins based on this just wrap the docker plugin binary.

zeyuye-airwallex:
@tboerger yeah i just realised (what you mean) and saw the below code in the `gcr/heroku/ecr` cmd

```go
	// invoke the base docker plugin binary
	cmd := exec.Command("drone-docker")
```

zeyuye-airwallex:
oh by the way, i forgot to mention, the reason we're using host level `/var/run/docker.sock` is to improve performance by caching the docker build content onto the host level. This way when next set of parallel job starts, it takes advantage of cached docker layer :)

PR #180, opened 2018-05-31
Add support for non label-schema labels

The-Loeki:
After the label-schema.org label support was added, this PR now adds the strongly suggested `org.label-schema.version` label by default and adds support for 'regular' labels similar to the tags and build args.
This incidently solves both comments in the original PR ;) 

For backwards compatibility I've kept the `label-schema` option.

# References
#134 
#135

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -172,6 +172,11 @@ func main() {
 			EnvVar: "PLUGIN_REPO",
 		},
 		cli.StringSliceFlag{
+			Name:   "Labels",
+			Usage:  "labels",
+			EnvVar: "PLUGIN_LABELS",
+		},
+		cli.StringSliceFlag{
 			Name:   "label-schema",
 			Usage:  "label-schema labels",
 			EnvVar: "PLUGIN_LABEL_SCHEMA",
@@ -242,6 +247,7 @@ func run(c *cli.Context) error {
 			Pull:        c.BoolT("pull-image"),
 			Compress:    c.Bool("compress"),
 			Repo:        c.String("repo"),
+			Labels:      c.StringSlice("labels"),
 			LabelSchema: c.StringSlice("label-schema"),
 			NoCache:     c.Bool("no-cache"),
 		},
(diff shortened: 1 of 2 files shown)


bradrydzewski:
`Labels` is a reserved keyword in the yaml, which means that `PLUGINS_LABELS` will not be populated and sent down to the plugin. Instead you should use a different name, such as `image_labels` in the yaml with the corresponding `PLUGINS_IMAGE_LABELS` environment variable.

cmd/drone-docker/main.go
```diff
@@ -171,6 +171,11 @@ func main() {
 			Usage:  "docker repository",
 			EnvVar: "PLUGIN_REPO",
 		},
+		cli.StringSliceFlag{
+			Name:   "Labels",
+			Usage:  "labels",
+			EnvVar: "PLUGIN_LABELS",
```
bradrydzewski: `labels` in the yaml is a reserved keyword, so we'll need to choose another attribute name

The-Loeki: I figure only `!label` is reserved?

bradrydzewski: container labels are reserved per the docker-compose specification. See https://docs.docker.com/compose/compose-file/#labels-1. Drone therefore uses these fields for special purposes

bradrydzewski: note that `dns` is also a reserved attribute, so you will notice it uses `custom_dns` with the `custom_` prefix. I would therefore recommend we follow the same pattern with `custom_labels`

The-Loeki: Is this what you had in mind ;)?

The-Loeki:
@bradrydzewski can you merge this or was there something else you wanted me to fix?

johnrengelman:
@bradrydzewski any additional feedback on this?
I'm looking at adding the build number label as we discussed at NeXT and figured I could get this across the finish line too.

bradrydzewski:
sorry for the delay, merged

PR #206, opened 2018-10-24
Cache from PR Cleanup

KayoticSully:
I started at the PR from (https://github.com/drone-plugins/drone-docker/pull/151) and integrated the general feedback on that PR. Originally, I had written this feature up as well but I want to give credit to @sethpollack since that PR was up a while ago.

My goal here is to get this change merged into master. I know it would be useful for my team and judging by some bump comments on the original PR it looks like many people could benefit.

Please let me know if there is any additional feedback, I'm motivated to drive this to completion asap.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -105,15 +105,6 @@ func (p Plugin) Exec() error {
 		fmt.Println("Registry credentials not provided. Guest mode enabled.")
 	}
 
-	// pre-pull cache image
-	for _, img := range p.Build.CacheFrom {
-		cmd := commandPull(img)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		trace(cmd)
-		cmd.Run()
-	}
-
 	if p.Build.Squash && !p.Daemon.Experimental {
 		fmt.Println("Squash build flag is only available when Docker deamon is started with experimental flag. Ignoring...")
 		p.Build.Squash = false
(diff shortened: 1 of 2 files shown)


shimunn:
Any reason why this ins't merged yet ?

docker.go
```diff
@@ -138,7 +144,9 @@ func (p Plugin) Exec() error {
 		trace(cmd)
 
 		err := cmd.Run()
-		if err != nil {
+		if err != nil && cmd.Args[1] == "pull" {
```
bradrydzewski: Can we place this code in a helper function (e.g. isPullCache) and check index size to prevent an index out of bound?

shimunn: this shouldn't go out of bounds as is, if you consider that Args[0] is docker and Args[1] has to be something at all times

bradrydzewski: unchecked bounds should be avoided, as should magic number of this sort. Moving this to a named function also makes the code more readable. Otherwise this code may make no sense to anyone that hasn't read this thread.

KayoticSully: My initial thinking was the same as @shimunn but your points are definitely valid and ultimately better for the reasons you stated.  Looks like @shimunn issued a PR against my branch in my fork with these changes. Ill get that resolved there so this can move forward

KayoticSully:
@bradrydzewski any chance of getting this merged? Are there more changes/feedback you'd like to see?

PR #207, opened 2018-10-26
Use the user specified Registry

benwilson512:
Hey! This PR ensures that the ecr executable honors the user specified registry.

The current behavior uses the registry returned by the ECR authentication token regardless of what the user specifies. While this registry is a good default, the token also authenticates the client to push to any other repository that has an AWS trust relationship established with the current account. We should let the user specify one such registry.

### DIFF
--- a/cmd/drone-docker-ecr/main.go
+++ b/cmd/drone-docker-ecr/main.go
@@ -19,11 +19,12 @@ const defaultRegion = "us-east-1"
 
 func main() {
 	var (
-		repo   = getenv("PLUGIN_REPO")
-		region = getenv("PLUGIN_REGION", "ECR_REGION", "AWS_REGION")
-		key    = getenv("PLUGIN_ACCESS_KEY", "ECR_ACCESS_KEY", "AWS_ACCESS_KEY_ID")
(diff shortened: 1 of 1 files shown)


benwilson512:
I was able to get this to build locally, but the docker image I built from the readme got:
```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```
when I changed my drone yaml to the new image. Quite possibly I simply messed up the image build.

benwilson512:
Are the instructions for building in the readme supposed to produce something that can be used in a `.drone.yml` file? I'd like to be able to test this out but I'm unclear on how to do so.

bradrydzewski:
if you use a custom tag (e.g. benwilson512/drone-docker) then you need to register your image with the drone server. Pass `DRONE_ESCALATE=benwilson512/drone-docker` to the Drone server. This will instruct the agent to run the build in privileged mode, allowing docker-in-docker.

benwilson512:
Ah excellent, thanks! Very new to the drone ecosystem still.

benwilson512:
I can confirm that this works.

tboerger:
Appveyor is expected to fail, so imho this is safe to merge.

PR #210, opened 2018-12-21
Returning full error message on doker login

jeroenrinzema:
Regarding: https://discourse.drone.io/t/unable-to-authenticate-plugins-docker-and-plugins-gcr/3309

This PR enables to return the full error message given by `docker login` to ease debugging.
Example error message:

```bash
time="2018-12-21T22:47:53Z" level=fatal msg="Error authenticating: Error response from daemon: Get https://gcr.io/v2/: unknown: Unable to parse json key.\n"
```

---

- Simplified docker login methods
- Piping password via stdin to avoid unnecessary error messages
- Return `docker login` error message if thrown instead of status code

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -1,6 +1,8 @@
 package docker
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


tboerger:
This feels pretty complicated.

PR #211, opened 2019-01-02
Update readme about plugins/docker has special privileges

kevinsimper:
I didn't know about this and I could not understand why I couldn't get other docker images to work the same way as plugins/docker worked. Wasted around 2 hours trying all kinds of stuff. Not mad, just want to spare other peoples time as I think it is quite an important undocumented feature, something quite unusual for Drone for what my impression has been.

As you can see it is not mentioned a single time in the docs, and being "container-native" you could assume that people would want to test building Docker images https://github.com/drone/docs/search?q=DRONE_ESCALATE&type=Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 Drone plugin to build and publish Docker images to a container registry.
 
+### Special privilieges
+
+This docker image is specified in Drone to run with Privilieged mode. If you fork this repository to run it on your own, you need to specify your docker image in the `DRONE_ESCALATE`, as the container needs to run in privilieged mode, otherwise it will not work! https://github.com/drone/drone/blob/master/cmd/drone-server/server.go
+
 ## Build
 
 Build the binary with the following commands:


README.md
```diff
@@ -2,6 +2,10 @@
 
 Drone plugin to build and publish Docker images to a container registry.
 
+### Special privilieges
+
+This docker image is specified in Drone to run with Privilieged mode. If you fork this repository to run it on your own, you need to specify your docker image in the `DRONE_ESCALE`, as the container needs to run in privilieged mode, otherwise it will not work! https://github.com/drone/drone/blob/master/cmd/drone-server/server.go
```
techknowlogick: TYPO, should be `DRONE_ESCALATE`

kevinsimper: Thank you for noticing 👍

README.md
```diff
@@ -2,6 +2,10 @@
 
 Drone plugin to build and publish Docker images to a container registry.
 
+### Special privilieges
+
+This docker image is specified in Drone to run with Privilieged mode. If you fork this repository to run it on your own, you need to specify your docker image in the `DRONE_ESCALATE`, as the container needs to run in privilieged mode, otherwise it will not work! https://github.com/drone/drone/blob/master/cmd/drone-server/server.go
```
tboerger: You got a typo in privileged multiple times, please drop the moving reference to the drone core source file, you should append "environment variable" to DRONE_ESCALATE.

Generally I'm not sure if the wording is that great.

kevinsimper: I can agree the wording can be better @tboerger :) How would it be said better?

kevinsimper:
@bradrydzewski I agree with you that hardcoding things is not so great idea, I am not sure that a Discourse thread is a better alternative since only the author can change the first post, unlike a Github Repo.

Since this plugin has special behavior which are undocumented, I think it is important to highlight it where most people would expect to find it. Either in the documentation or the plugin that has the special behavior.

Or maybe remove the special behavior so it doesn't have to be documented and treat all containers equal in a container native ecosystem where side-effects are unwanted :)

Or a mark in the UI showing that certain containers are treated differently than others because of a string match in their name :)

Just some ideas :)

bradrydzewski:
The reason I suggest discourse is that the instructions can be augmented with comments, as opposed to a pull request. I do not want version-specific information to leak into the plugin dogumentation, so it needs to be documented somewhere else.

Changing Drone behavior is not an option for security reasons. We whitelist specific plugins so that you do not need to enable trusted mode for your entire repository to use this plugin, which would reduce security for public github repositories and allow a malicious pull request to pwn the host machine.

kevinsimper:
> The reason I suggest discourse is that the instructions can be augmented with comments, as opposed to a pull request. I do not want version-specific information to leak into the plugin dogumentation, so it needs to be documented somewhere else.

Yes, that makes sense if they refer to some original documentation. Forums are seldom a good documentation resource to scavenge through.

> Changing Drone behavior is not an option for security reasons. We whitelist specific plugins so that you do not need to enable trusted mode for your entire repository to use this plugin, which would otherwise reduce security for public github repositories and allow a malicious pull request to pwn the host machine.

Yeah I understand why, what do you think of my other suggestions that does not change the behavior? @bradrydzewski :)

bradrydzewski:
I think a link to documentation / thread is sufficient to solve this problem, as opposed to development changes to the user interface which are costly to develop and maintain.

tboerger:
Thanks for the PR, I have added a generic note to the README at https://github.com/drone-plugins/drone-docker#usage now.

kevinsimper:
@tboerger That is a pretty good description, good edit!

PR #227, opened 2019-03-07
Update docker base image on Linux

dschmidt:
Don't ask, @tboerger  :-D


codestation:
This is required to be able to use the new buildkit feature from 18.09.

dschmidt:
@codestation Yeah, exactly - that's why I sent it :-)

tboerger:
You can use `plugins/docker:latest`, `plugins/docker:18.09.0`, `plugins/docker:18.09`, `plugins/docker:18` on Linux with the architecture amd64, arm64v8 and arm32v7 now.

Issue #191, opened 2018-09-10
Azure CR plugin


jpds:
It would be neat if this plugin had integration for Azure Container Registry.

There is some example code for the Azure SDK at: https://github.com/Azure-Samples/azure-sdk-for-go-samples/pull/215

jpds:
I managed to get this working with plain 'headless' docker and a service principal per the Microsoft documentation:

- https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-service-principal

bradrydzewski:
This is not something I plan to work on, but would be happy to accept a pull request.

PR #228, opened 2019-03-26
Create ACR plugin extension

Lucretius:
Resolves https://github.com/drone-plugins/drone-docker/issues/191

This does essentially the same thing as the GCR plugin - just a minor extension on top of existing drone-docker which allows setting of `service_principal_client_id` and `service_principal_client_secret`.  Note that these can also be used as the `username` and `password` without issue.  In the case that `admin` is enabled for an Azure Container Registry, the admin username and password can simply be used for the `username` and `password` fields.

I have tested out the linux build with:
1) setting the `service_principal_client_id` and `service_principal_client_secret` which worked
2) not explicitly setting the registry
3) setting a simple admin username and password using the standard drone-docker `username` and `password` fields.

Again, the only thing this does is add the setting of two more properties: `service_principal_client_id` and `service_principal_client_secret`, which are used in place of the standard `username` and `password`.  All of the build steps I simply copied from the `gcr` setup.


docker/acr/manifest.tmpl
```diff
@@ -0,0 +1,37 @@
+image: plugins/acr:{{#if build.tag}}{{trimPrefix build.tag "v"}}{{else}}latest{{/if}}
```
tboerger: You got to change `{{trimPrefix build.tag "v"}}` to `{{trimPrefix "v" build.tag}}`, same for the other lines.

Lucretius:
Thanks @tboerger - I made the changes in the manifest file for `acr`.  Would you like me to change the rest of the manifest files to match?  I assume they will suffer from the same issue.

tboerger:
For the other manifests I have merged a pr today.

PR #233, opened 2019-04-24
Add Assume Role Option For ECR Pushes

jwalton9:
This PR adds a new option to the ECR plugin to assume an IAM role to use with the ECR client, rather than only allowing using a single role on a single account

This will allow multi account use cases where there may not be push access between the AWS user and the ECR registry but the user is able to assume a role that has push access.

E.G. A user in a platform account that runs in CI trying to push to an ECR in another account that only has push access to a single role in that account


mjschultz:
@bradrydzewski what is the criteria for getting this merged in?

tboerger:
It requires an additional review by the maintainers team, not focused on brad.

cmd/drone-ecr/main.go
```diff
@@ -42,12 +44,12 @@ func main() {
 	}
 
 	sess, err := session.NewSession(&aws.Config{Region: &region})
-
+
```
lafriks:
```suggestion

```

PR #240, opened 2019-07-15
Allow to cache the intermediate steps of a multi-stage build

codestation:
This PR allows to cache the target image of a multi-stage build so it can be pulled later as a cache and speed up builds. Since `docker build` has to be run twice with the same context, i added support to  `stream` the context and save time by don't sending it twice to the daemon.

The patch is simple enough and i have been using for two months without a problem with multiple deployments and allowed me to simplify my drone config files. This approach would be more simpler than the alternative of declaring multiple steps on the `.drone.yml` file.

```yaml
steps:
  - name: build
    image: codestation/drone-docker
    settings:
      repo: registry.example.com/someimage
      registry: registry.example.com
      tags: latest
      stream: true
      target: builder
      push_target: true
      experimental: true
      cache_from:
        - registry.example.com/someimage:builder
        - registry.example.com/someimage:latest
      username:
        from_secret: docker_username
      password:
        from_secret: docker_password
```

The build will also work without `stream` (that option requires `experimental`) but it will be a little slower because it has to send the context twice.

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -248,6 +248,11 @@ func main() {
 			Usage:  "additional host:IP mapping",
 			EnvVar: "PLUGIN_ADD_HOST",
 		},
+		cli.BoolFlag{
+			Name:   "stream",
+			Usage:  "stream the build context",
+			EnvVar: "PLUGIN_STREAM",
+		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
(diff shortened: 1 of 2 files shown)


tboerger:
Not sure if this should directly get into this important plugin. @bradrydzewski what do you think?

sebasptsch:
Is there an implemented alternative to this pull request seeming as it's taking a while to merge into prod?

PR #241, opened 2019-07-16
Add shm-size

jpic:


docker.go
```diff
@@ -221,6 +222,9 @@ func commandBuild(build Build) *exec.Cmd {
 	if build.NoCache {
 		args = append(args, "--no-cache")
 	}
+	if build.ShmSize != "" {
+		args = append(args, "--shm-size=" + build.ShmSize)
```
tboerger:
```diff
-		args = append(args, "--shm-size=" + build.ShmSize)
+		args = append(args, "--shm-size", build.ShmSize)
```

jpic: Thanks for your review !

Pushed your fix

jpic:
Hi all,

So, no release for this ?

Thanks

cmd/drone-docker/main.go
```diff
@@ -223,6 +223,10 @@ func main() {
 			Usage:  "do not use cached intermediate containers",
 			EnvVar: "PLUGIN_NO_CACHE",
 		},
+		cli.StringFlag{
+			Name:   "shm-size",
+			Usage:  "set the shm-size",
+			EnvVar: "PLUGIN_SHM_SIZE",
```
tboerger: Missing the closing `},`

tboerger:
@bradrydzewski fine to merge?

cmd/drone-docker/main.go
jpic: Thank you @tboerger

samtgarson:
Could we get this merged? Would be very useful.

Alternatively @jpic @tboerger do either of you have a built image of this I could use in the meantime?

jpic:
I currently have that one
```
steps:    
- name: build    
  image: yourlabs/docker    
  privileged: true    
  volumes:    
  - name: docker    
    path: /var/run/docker.sock    
  commands:    
  - echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin
  - docker build    
        --shm-size 512M    
        -t $REPO:$DRONE_COMMIT    
        --build-arg GIT_COMMIT=$DRONE_COMMIT    
```

psheth-twist:
Can we please get this merged to master?

ashwilliams1:
The shm_size is only available to trusted pipelines. I am not sure this change can be merged since it would circumvent this policy and allow untrusted pipelines to set shm_size.  It would require further analysis to understand the implications.

In the meantime, we recommend building the image from source with this patch.

PR #242, opened 2019-07-24
append tags with autotag instead of replacing

### LINKED ISSUE #201: When auto_tag is specified, tags is ignored
Let's imagine we have the following pipeline : 

```
pipeline:
  push-image:
    image: plugin/docker
    auto_tag: true
    tags:
      - ${DRONE_COMMIT_SHA}
```

When auto_tag is set, only the automatic tags will be set, ${DRONE_COMMIT_SHA} will be ignored.

Wouldn't it be better for them to be merged ?

foosinn:
As in #201 mentioned it can be helpful to allow static tags in addition to the auto tags. I. e.: having a production tag that always matches the last tag.

closes #201

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -280,10 +280,10 @@ func run(c *cli.Context) error {
 			c.String("commit.ref"),
 			c.String("repo.branch"),
 		) {
-			plugin.Build.Tags = docker.DefaultTagSuffix(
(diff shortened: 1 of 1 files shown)


tboerger:
@bradrydzewski since this impacts all existing docker configs, wdyt?

bradrydzewski:
I would guess most people are using either `auto_tag` or `tags` but not both, so the impact of this change should be very minimal. I am neutral to the change itself, and am fine with merging.

bradrydzewski:
I reverted this due to reports that it broke the current image

PR #243, opened 2019-08-06
Revert "append tags with autotag instead of replacing"

bradrydzewski:
Reverts drone-plugins/drone-docker#242


tboerger:
Who adds this feature again? It didn't caused the issue :)

bradrydzewski:
yes, sorry, just to clarify the problem was not with this pull request, however, we rolled it back because immediately after merging we had reports of an issue. It turns out the issue was related to the `docker:dind` image which we resolved by hard-coding the last-known working version of that image.

PR #248, opened 2019-08-15
ECR: Respect the registry option

mjschultz:
This is just https://github.com/drone-plugins/drone-docker/pull/207 with a new merge of master. It would be nice to get this in master.


tboerger:
Why are you opening a new PR for that?

mjschultz:
Because I can't commit to the other.

PR #251, opened 2019-09-09
Remove secrets from build arg trace

nodefortytwo:
--build-args often contain secrets which drone this plugin prints.

This change does two things:
- makes trace() testable
- adds tests for trace
- removes the values from any --build-arg

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -20,7 +20,7 @@ func (p Plugin) startDaemon() {
 		cmd.Stderr = ioutil.Discard
 	}
 	go func() {
-		trace(cmd)
+		trace(os.Stdout, cmd)
 		cmd.Run()
 	}()
 }
--- a/docker.go
+++ b/docker.go
@@ -2,8 +2,10 @@ package docker
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 	"time"
 )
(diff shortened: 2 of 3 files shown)


tboerger:
But the drone server is replacing secrets by asterisks, so why is this still needed?

nodefortytwo:
Is that supposed to happen on 0.8? certainly haven't seen that behaviour.

tboerger:
I can't remember the behavior on 0.8, but since 0.8 doesn't receive any updates anymore you should really upgrade to 1.x.x

nodefortytwo:
I understand that however migrating our org of 100s of devs, 2000+ repos and a large ecosystem of 0.8 tooling its not easy.

Given that 0.8 support has been dropped is it better for us to just publish and maintain a fork of this plugin?

tboerger:
For the sake of simplicity I would prefer that, yeah. But I don't know @bradrydzewski's opinion on that.

bradrydzewski:
while I do very much appreciate the pull request, as mentioned, this is solved in 1.0.0 and higher with automatic masking of secrets in logs. 0.8 is also end-of-life. For this reason we recommend either upgrading or maintaining a fork of the plugin.

espears1:
@bradrydzewski can you confirm if `build_args` entries will be masked in drone 1, or if you specifically need `build_args_from_env` and to reference `from_secret` entries of `environment`? I ask because I am using a setup with

```
build_args:
  from_secret: my-drone-build-args-secret
```

where `my-drone-build-args-secret` is a large, bundled `FOO=bar,BAZ=wah` combined string stored in drone secrets, but it still shows up unmasked in plain text in the drone 1.0 logs.

I'm trying to switch to the `build_args_from_env` option on the hopes it will mask the secret env vars, but I can't find any drone 1.0 docs that actually explicitly confirm this.

PR #252, opened 2019-10-10
Revert "Reapply append tags with autotag instead of replacing"

tboerger:
Reverts drone-plugins/drone-docker#245. @foosinn if you still would like to see that feature please make sure that auto tagging won't get broken as with this patch it always applies `latest`.


tboerger:
Merging this as the current latest version is broken with auto_tag enabled.

PR #257, opened 2019-12-11
Add --network

recamshak:
Add the `--network` flag during build


bradrydzewski:
I am not sure this makes sense to enable for this plugin.  It runs docker-in-docker which means --network will not have any practical impact since everything is happening inside the plugin container, which itself is running inside an isolated network.

recamshak:
I thought this solved one of my issues. But after further experiments it seems it was not related.
As you said this probably has no real effect. I'll close this PR.

PR #260, opened 2020-01-09
Add DinD proxy support

xvzf:
When running DroneCI behind a corporate Proxy, it might be necessary to pass proxy settings to the docker daemon.

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -7,7 +7,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
-	"github.com/drone-plugins/drone-docker"
+	docker "github.com/drone-plugins/drone-docker"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


bradrydzewski:
I am not sure we need explicit fields for this since you can add these parameters directly to the environment section (below). You also have the option to set these parameters globally with the runner.

```
environment:
  HTTP_PROXY: ...
  HTTPS_PROXY: ...
```

xvzf:
Can you confirm this is working? I tried it and had no success.

e.g. this does not work:
```yaml
- name: build_dockerimage
  image: plugins/docker
  settings:
    username:
      from_secret: reg_username
    password:
      from_secret: reg_password
    dockerfile: Dockerfile.kubeagent
    repo:
      from_secret: reg_repo_kubeagent
    registry:
      from_secret: reg_registry
    auto_tag: true
    build_args:
      - HTTP_PROXY=http://proxy.cluster.local
      - HTTPS_PROXY=http://proxy.cluster.local
    environment:
      HTTP_PROXY: http://proxy.cluster.local:80
      HTTPS_PROXY: http://proxy.cluster.local:80
```

xvzf:
Okay, global environment is not working, and I have to indent correctly

PR #266, opened 2020-03-12
adding windows 1909

luthermonson:
just copy pasta for 1909, not sure it will work but here we go.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -117,6 +117,44 @@ depends_on:
 
 ---
 kind: pipeline
+type: ssh
+name: windows-1909-amd64-docker
+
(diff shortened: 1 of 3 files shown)


tboerger:
This won't work as we don't have such worker yet.

luthermonson:
@tboerger bummer.. going to get it sometime?

tboerger:
Waiting for @bradrydzewski to provide them

luthermonson:
oh great, so leaving this up then and we'll see what happens when we get some builders. thanks!

cjellick:
any movement on this?

PR #268, opened 2020-03-24
Added AuthConfig to login to multiple  registry

Koma-Andrea:
I've added the possibility to create a docker config for the root user from variable.
In this way using a secret you will be able to add all the authentication parameters for all the registries that you want.

In this example I need to authenticate to
-  registry.redhat.io (as build step)
- quey.io (as Dockerfile FROM as sourcedata)
- docker.io (as Dockerfile FROM as dockerexample)
- selfhosted_registry (as Dockerfile FROM as localdata)

 example:
```yaml
kind: pipeline
name: default
image_pull_secrets:
  - dockerconfig
steps:
  - name: pull ubi7
    image: registry.redhat.io/ubi7/ubi:7.7
    purge: false
    commands:
      - echo $HOSTNAME

  - name: pull ose-ansible
    image: registry.redhat.io/openshift3/ose-ansible:latest
    purge: false
    commands:
      - echo $HOSTNAME

  - name: build and publish on internal registry
    # image: plugins/docker
    # (ATM is in my repo of you want to try this functionality)
    image: docker.io/koma85/plugins-docker 
    privileged: true
    settings:
      dry_run: false
      dockerfile: Dockerfile
      purge: false
      auto_tag: true
      registry: docker.io
      repo: koma85/acontainerthatiwanttobuild
      authconfig:
        from_secret: dockerconfig
```

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -208,6 +208,11 @@ func main() {
 			Usage:  "docker email",
 			EnvVar: "PLUGIN_EMAIL,DOCKER_EMAIL",
 		},
+		cli.StringFlag{
+			Name:   "docker.authconfig",
+			Usage:  "docker json authconfig content",
+			EnvVar: "PLUGIN_AUTHCONFIG,DOCKER_AUTHCONFIG",
+		},
 		cli.BoolTFlag{
 			Name:   "docker.purge",
 			Usage:  "docker should cleanup images",
@@ -240,10 +245,11 @@ func run(c *cli.Context) error {
 		Dryrun:  c.Bool("dry-run"),
 		Cleanup: c.BoolT("docker.purge"),
 		Login: docker.Login{
-			Registry: c.String("docker.registry"),
-			Username: c.String("docker.username"),
-			Password: c.String("docker.password"),
-			Email:    c.String("docker.email"),
+			Registry:   c.String("docker.registry"),
+			Username:   c.String("docker.username"),
+			Password:   c.String("docker.password"),
+			Email:      c.String("docker.email"),
+			AuthConfig: c.String("docker.authconfig"),
 		},
 		Build: docker.Build{
 			Remote:      c.String("remote.url"),
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -209,9 +209,9 @@ func main() {
 			EnvVar: "PLUGIN_EMAIL,DOCKER_EMAIL",
 		},
 		cli.StringFlag{
-			Name:   "docker.authconfig",
-			Usage:  "docker json authconfig content",
-			EnvVar: "PLUGIN_AUTHCONFIG,DOCKER_AUTHCONFIG",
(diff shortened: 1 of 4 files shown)


docker.go
```diff
@@ -83,6 +84,35 @@ func (p Plugin) Exec() error {
...
+			fmt.Println("Error setting permissions on root's docker auth config file: ")
+			fmt.Println(err_chmod)
+		}
+		_, err_str := cfile.WriteString(p.Login.AuthConfig)
+		if err_str != nil {
+			fmt.Println("Error filling root's docker auth config file: ")
+			fmt.Println(err_str)
+		} else {
+			cfile.Close()
+		}
```
bradrydzewski: we can combine these three commands into a single command:

```
ioutil.WriteFile("/root/.docker/config.json", []byte(p.Login.AuthConfig, 0600)
```

cmd/drone-docker/main.go
```diff
@@ -208,6 +208,11 @@ func main() {
 			Usage:  "docker email",
 			EnvVar: "PLUGIN_EMAIL,DOCKER_EMAIL",
 		},
+		cli.StringFlag{
+			Name:   "docker.authconfig",
+			Usage:  "docker json authconfig content",
+			EnvVar: "PLUGIN_AUTHCONFIG,DOCKER_AUTHCONFIG",
```
bradrydzewski: this is more of a small nit, but let's name this `PLUGIN_CONFIG` since technically speaking it is a config file.  The config file includes auth, but it could technically contain other non-auth related configuration parameters.

docker.go
```diff
@@ -83,6 +84,35 @@ func (p Plugin) Exec() error {
 		time.Sleep(time.Second * 1)
 	}
 
+	// Create Auth Config File
+	if p.Login.AuthConfig != "" {
+		fmt.Println("AuthConfig provided.")
+		err_mkdir := os.MkdirAll("/root/.docker", 0600)
```
bradrydzewski: The file paths and names should be stored in `daemon.go` and `daemon_win.go` as constants. This will ensure the correct path is for windows. I am not quite sure which path we should use for windows.  cc @donny-dont ?

bradrydzewski: edit: for windows the path is `C:\ProgramData\Docker\config\daemon.json`

bradrydzewski:
Thanks for sending the pull request.  A few minor changes requested and then we can merge.

docker.go
Koma-Andrea: Perfect I'll test it immediatly

cmd/drone-docker/main.go
Koma-Andrea: Good Idea

docker.go
Koma-Andrea: Ok, but what is the default path for windows?

bradrydzewski: on windows it is `C:\ProgramData\Docker\config\daemon.json`

Koma-Andrea:
Made requested changes. Opening new PR

PR #270, opened 2020-03-30
Add multiple insecure registries support

orangedeng:
Add flag daemon.insecure-registry to indecate multiple insecure registry.
Append insecure registry parameters when starting daemon.


tboerger:
There is already an insecure flag.

eliandoran:
Hello, @tboerger .

I have come across exactly across a scenario where I needed this PR instead of the `insecure` flag.

Consider the following scenario:

- We have two different Docker repositories, both insecure (run on a local network only): orange.example.com and pear.example.com.
- The docker image that we want to build needs an image from the insecure orange.example.com.
- The docker image that we want to build has to be published to the insecure pear.example.com.

As such, the Dockerfile contains:
```Dockerfile
FROM orange.example.com/my-secret-base-image
```

The build step in `.drone.yml` contains:
```
- name: deploy-tag
  image: "drone-docker"
    settings:
      repo: "pear.example.com/${DRONE_REPO_NAME}"
      registry: "pear.example.com"
      insecure: true
```

On a first glance, this should work: we want to publish to pear.example.com which is marked with the `insecure` flag. Upon publishing you get:

```
Get "https://orange.example.com/v2/": dial tcp 1.2.3.4:443: connect: connection refused
```

The reason? The `insecure` flag only adds the following `--insecure-registry` flag:

```
/usr/local/bin/dockerd --data-root /var/lib/docker --host=unix:///var/run/docker.sock --insecure-registry pear.doran.xyz
```

Basically, the `insecure` flag would only add the registry we are deploying to, but not any registries our image might depend on.

With this PR, I was able to fix this issue by changing the `.drone.yml` as follows:

```
- name: deploy-tag
  image: "drone-docker"
    settings:
      repo: "pear.example.com/${DRONE_REPO_NAME}"
      registry: "pear.example.com"
      insecure_registry: [ "orange.example.com" ]
      insecure: true
```

This will cause the build to succeed:

```
/usr/local/bin/dockerd --data-root /var/lib/docker --host=unix:///var/run/docker.sock --insecure-registry pear.doran.xyz --insecure-registry orange.doran.xyz
```

Personally, I think this is a very useful addition to the plugin as I came across a specific use case for this. However, I would understand that it is a less frequent use case.

PR #271, opened 2020-04-21
Add a flag for "normalizing" docker tags to contain only valid characters

decarboxy:
We sometimes use git branch names as docker tag names.  git branch names can have / characters, while docker tags cannot.  This PR adds a new boolean option `tags.normalize` which replaces all ascii characters that cannot exist in a docker tag with `-`. The flag is off by default.


tboerger:
We could add the normalization to the plugin code, but not as part of the main function or main function file.

PR #274, opened 2020-04-25
Add quiet build option

gary-kim:
As title.
This adds the option for using the `--quiet` flag during builds.


gary-kim:
Sorry, just noticed a tiny typo.

Issue #237, opened 2019-06-26
add auto_tags depth option


thomasf:
We release new versions of multiple minor version images in some builds which means that the `{image}:{major}`  tag randomly switches depending on the last build of any 1.x release series . 

So either `auto_tag` can transform into in addition of being an boolean  option also support a list of version tags or it's own option:

```yml
auto_tag_versions: ['major', 'minor', 'patch']
```

bradrydzewski:
auto_tag is meant to be opinionated and simple. Either you use it as-is or you do not use it at all. There are too many ways to customize tags and if we merge this, we have to merge other requests to customize auto_tag (add branch, commit sha, etc). I think a good compromise would be to allow custom tag logic via an external starlark script that you checkin to your repository. Thoughts @tboerger ?

thomasf:
We also have a related problem with several repositories where we release multiple services from one repo using tag prefixes.

Auto tag cannot handle it tags like `cmd/foobar/1.4.3` or `cmd/foobar/v1.4.3`.

The `.drone.yml` section looks something like this.

```yml
- name: docker/foobar
  image: plugins/docker
  settings:
    repo: dreg.example.com/project/foobar
    dockerfile: contrib/drone/Dockerfile.foobar
  when:
    ref:
    - refs/tags/cmd/foobar/**
```

I could also write a separate drone plugin that can create `.tags` files based on various rules.. Right now each project has some code to handle these things individually but scripting support build into this image would be much more convenient than any other solution.

thomasf:
I could maybe have a go at it experimenting some.. I have been wanting to try use *lark for something but have not had a real project.. I'd guess that we would add in bindings to some semverparsing go code as well so there are a reasonable set of utility functions to work with. Does any other drone plugin has lark scripting yet?

tboerger:
I'm not a big fan of changing the auto tag behavior, as brad said it's pretty simple intentionally.

Every other special case can make use of the tags file.

From my experience your use case seem to be more of a corner case.

thomasf:
supporting multiple minor releases of software is certainly not unusual

thomasf:
Btw. my second example more or less maps to how go modules works for multi module repositories as well  https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories  From my experience it's the go to solution for when you want a lot of projects in one synchronized code base with git.

gypapp:
[...]

PR #278, opened 2020-05-15
DefaultTags() returns non-semver tags as-is.

gypapp:
The previous behavior of `DefaultTags()` causes problems to us because our staging environment depends on (ie. pulls) the `latest` images so whenever not a semver tag hits the repository, `auto_tag` eventually moves the `latest` tag to another arbitrary image in the repository.

IMHO it would be better to make the build fail or use the provided tag as is.

For further details visit: https://github.com/drone-plugins/drone-docker/issues/237#issuecomment-627393512

### DIFF
--- a/tags.go
+++ b/tags.go
@@ -43,7 +43,7 @@ func DefaultTags(ref string) []string {
 	v := stripTagPrefix(ref)
 	version, err := semver.NewVersion(v)
 	if err != nil {
-		return []string{"latest"}
+		return []string{v}
 	}
 	if version.PreRelease != "" || version.Metadata != "" {
 		return []string{
--- a/tags_test.go
+++ b/tags_test.go
@@ -35,8 +35,11 @@ func TestDefaultTags(t *testing.T) {
 		{"refs/tags/v1.0.0", []string{"1", "1.0", "1.0.0"}},
 		{"refs/tags/v1.0.0-alpha.1", []string{"1.0.0-alpha.1"}},
 
+		// not semver tags as-is
+		{"refs/tags/x1.0.0", []string{"x1.0.0"}},
+		{"refs/tags/local_test", []string{"local_test"}},
+
 		// malformed or errors
-		{"refs/tags/x1.0.0", []string{"latest"}},
 		{"v1.0.0", []string{"latest"}},
 	}
 


tboerger:
IMHO it should fail if it's not a semver tag. For cases without a semver tag somebody should not use auto tagging.

gypapp:
@tboerger you must better know how auto-tagging and semantic versioning are supposed to work in drone-docker. However, I would rather like to see the semver support as an optional feature on top of auto-tagging because in this case auto-tagging could be used in more scenarios (like mine). As it stands now, I nearly duplicate all the functionalities of auto-tagging in my drone pipeline configuration just because of that issue.

This was the reason why, as I stated in the description, I followed @bradrydzewski comment in this pull request.

tboerger:
I shared my opinion. IMHO the advantage of the autotagging is the automated split of the versions. For taking the tag as it is nobody needs autotagging, just use the drone env variables like they are.

bradrydzewski:
> IMHO it would be better to make the build fail or use the provided tag as is.
> IMHO it should fail if it's not a semver tag

I agree with both of you that ideally it would fail (or just skip the step similar to how we handle pull requests) if semver parsing failed. The reason I suggested using the tag as-is instead of failing is because I was concerned failing or skipping could inadverently break existing pipelines that do not expect this behavior -- although perhaps this is not a valid concern.

Either way I agree we need to change the logic, either as implemented in this pull request, or by failing or skipping.

tboerger:
Just taking the tag like it is would also be some kind of breaking change... It would silently build another tag as the people out there might already expect ;)

tboerger:
@bradrydzewski if you are fine with this change go ahead and merge it.

gypapp:
I would be more than happy if you merge it.

However, if you think I could add a `tags.semver` boolean option to the `cli.Context` to be more backward-compatible and provide some control. It would be `true` by default meaning it skips the entire build (`return nil` before calling `plugin.Exec()` - I suppose this is what @bradrydzewski meant by skipping). Otherwise, it would enable us to use the tag as is.

bradrydzewski:
I appreciate the option, but would prefer we avoid adding more configuration parameters to the plugin. Since any change we make is going to be breaking, perhaps we just bite the bullet and modify the behavior to fail the plugin if semver parsing fails. Draft of this change at  https://github.com/drone-plugins/drone-docker/pull/279.

PR #280, opened 2020-05-21
compose DRONE_COMMIT_SHA and DRONE_STAGE_NUMBER as docker temporary tag.

18981119465:
When we use drone-plugins/drone-docker in a multi-stage drone pipeline for multi-module maven project, the pipeline failed to build container image. After investigating the cause of the failure, we found that the same docker temporary tag was used in every stage for the drone-docker plugin. So, we compose DRONE_COMMIT_SHA and DRONE_STAGE_NUMBER as docker temporary tag to solve this problem.

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -238,6 +238,12 @@ func main() {
 			Usage:  "additional host:IP mapping",
 			EnvVar: "PLUGIN_ADD_HOST",
 		},
+		cli.StringFlag{
+			Name:   "stage.number",
+			Usage:  "drone stage number",
+			EnvVar: "DRONE_STAGE_NUMBER",
+			Value:  "0",
+		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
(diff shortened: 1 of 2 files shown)


klamouri:
This is something we also had issues with. We would appreciate it if it could be reviewed and released :)

tboerger:
This can happen only if you are using it with an unofficial not really supported way, I guess you are using it with a host docker socket, otherwise such issues can't happen.

18981119465:
> This can happen only if you are using it with an unofficial not really supported way, I guess you are using it with a host docker socket, otherwise such issues can't happen.

Yes, we are using it with a host docker socket and tagged the docker image to our local docker registry to avoid bad networks to hub.docker.com. Isn't this a typical usage scenario for drone-docker plugin?

klamouri:
@tboerger We do mount the host docker socket. We do this to share the layer cache and speed up builds if the worker already build that image in the past. Any work around to not mount the host socket but still try to use the host layer cache?

tboerger:
IMHO it's not the task of the plugin to build temporary tags that work outside of the official supported way. You can always construct your own temporary tag which is safe for your environment.

But maybe @bradrydzewski got a different opinion?

PR #281, opened 2020-05-22
Adding Ignore Checks for prune and rmi

luthermonson:
We are trying to utilize image caching from the host machine a bit more and we're using the docker socket volume mount and seeing an issue with the `docker prune` such as the following:

pipeline is here : https://github.com/rancher/rancher/blob/master/.drone.yml#L345-L452
error: https://drone-publish.rancher.io/rancher/rancher/2828/3/4 pertinent parts below.
```
215 | + C:\bin\docker.exe rmi 2cd59a6e1694658e28b85deba6c9440f7496daf0
216 | Untagged: 2cd59a6e1694658e28b85deba6c9440f7496daf0:latest
217 | + C:\bin\docker.exe system prune -f
218 | Error response from daemon: a prune operation is already running
219 | exit status 1
```

This PR adds a similar check as the `isCommandPull` check so that we ignore the error and move on and assume the next job will successfully call `rmi` and `prune`.

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -135,6 +135,10 @@ func (p Plugin) Exec() error {
 		err := cmd.Run()
 		if err != nil && isCommandPull(cmd.Args) {
 			fmt.Printf("Could not pull cache-from image %s. Ignoring...\n", cmd.Args[2])
+		} else if err != nil && isCommandPrune(cmd.Args) {
+			fmt.Printf("Could not prune system containers. Ignoring...\n")
+		} else if err != nil && isCommandRmi(cmd.Args) {
+			fmt.Printf("Could not remove image %s. Ignoring...\n", cmd.Args[2])
 		} else if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


bradrydzewski:
If you are mounting the host machine docker socket you should should set `purge: false` to prevent the system from running prune commands.

luthermonson:
@bradrydzewski i want the purge of the build hash container and the prune is fine to run just not at the same time. i want purge set to true and utilize the host image cache as we pull ubuntu, servercore etc. a million times in the plugins/docker container without the mounted socket/pipe.

bradrydzewski:
got it.  I think these changes are fine since this plugin primarily targets docker-in-docker. A failed prune inside docker-in-docker is not a huge problem since everything is ephemeral anyway.  I think we are good to merge.

luthermonson:
@bradrydzewski thanks! ill watch for a tag

PR #283, opened 2020-05-27
Add settings nethost for build docker image

doubles-ss:
I use drone in kubernetes and found some problem with network. When I use this plugins to build docker image it struck when download package from internet, but other steps are work well.
After investigate found that if add option --network=host to use hostnetwork (POD) it similar to other steps environment.


tboerger:
There are unrelated changes included, you got to rebase properly before I'm able to give a real review.

ashwilliams1:
I would be concerned with the security impact of such a change on the docker runner

doubles-ss:
> There are unrelated changes included, you got to rebase properly before I'm able to give a real review.

Sorry for the accident. I revert to my original purpose.

doubles-ss:
> I would be concerned with the security impact of such a change on the docker runner

@ashwilliams1 after investigating with docker runner.
docker-runner: create containers to run each step.
and for this image, it creates a container for this step and runs dockerd inside it.
and my purpose is when we run docker build inside this step, I need option to use network=host for build time that means use the network of this container which docker-runner creates for this step just like other steps use instead of a new network that DinD created.

tboerger:
Codewise it looks good to me, but not sure about the security impact. @bradrydzewski your opinion?

bradrydzewski:
I am not sure about the security impact either, but that is something we should know before merging.  In the short term, you can always provide the runner with your own custom build.  You just need to pass `DRONE_RUNNER_PRIVILEGED_IMAGES=double-ss/docker` (or whatever you name your alternate image) to the runner.

PR #295, opened 2020-08-04
Allow image loading instead of building

allgreed:
This is an answer to the following topic: https://discourse.drone.io/t/how-to-push-a-tar-gz-docker-image-on-drone-cloud/7749
Hope that having an ability to easily use Nix with Drone Cloud benefit both communities.

Questions:
- is this a sensible solution to the problem at hand?
- what is the "proper" naming convention and grouping for CLI arguments? IMO it didn't fit `Deamon` nor `Build` struct yet I don't feel very comfortable expanding the main `Plugin` struct
- is it possible to extract information from executed commands? I'd like to capture the loaded image tag and use it in further plugin execution, but not sure if this is sensibly doable with the command pattern architecture of the plugin.

The current ad hoc proposed API is:
```yaml
load_image: true
load_image_location: ./foo/bar.tar.gz
load_image_tag: my-builded-image:latest # this is the tag of the zipped image, required for re-tagging it once it's loaded
```

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -238,6 +238,21 @@ func main() {
 			Usage:  "additional host:IP mapping",
 			EnvVar: "PLUGIN_ADD_HOST",
 		},
+		cli.BoolFlag{
+			Name:   "load-image",
+			Usage:  "load image from a file instead of building it",
+			EnvVar: "PLUGIN_LOAD_IMAGE",
+		},
+		cli.StringFlag{
+			Name:   "loaded-image-tag",
+			Usage:  "the tag of the loaded image",
+			EnvVar: "PLUGIN_LOAD_IMAGE_TAG",
+		},
+		cli.StringFlag{
+			Name:   "loaded-image-location",
+			Usage:  "the location of image file",
+			EnvVar: "PLUGIN_LOAD_IMAGE_LOCATION",
+		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
@@ -247,7 +262,10 @@ func main() {
 
 func run(c *cli.Context) error {
 	plugin := docker.Plugin{
-		Dryrun:  c.Bool("dry-run"),
+		Dryrun:            c.Bool("dry-run"),
+		Loadimage:         c.Bool("load-image"),
+		Loadedimagetag:    c.String("loaded-image-tag"),
+		Loadedimagelocation: c.String("loaded-image-location"),
 		Cleanup: c.BoolT("docker.purge"),
 		Login: docker.Login{
 			Registry: c.String("docker.registry"),
--- a/docker.go
+++ b/docker.go
@@ -61,11 +61,15 @@ type (
 
 	// Plugin defines the Docker plugin parameters.
 	Plugin struct {
-		Login   Login  // Docker login configuration
-		Build   Build  // Docker build configuration
-		Daemon  Daemon // Docker daemon configuration
-		Dryrun  bool   // Docker push is skipped
-		Cleanup bool   // Docker purge is enabled
+		Login   Login         // Docker login configuration
+		Build   Build         // Docker build configuration
+		Daemon  Daemon        // Docker daemon configuration
+		Dryrun  bool          // Docker push is skipped
+		Cleanup bool          // Docker purge is enabled
+		Loadimage bool        // Docker load image is enabled
+		// TODO: extract this from the loading and remove references
+		Loadedimagetag string // the tag of the loaded image
+		Loadedimagelocation string // the location of image file
 	}
 )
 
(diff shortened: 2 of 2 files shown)


cmd/drone-docker/main.go
```diff
@@ -238,6 +238,21 @@ func main() {
...
 		},
+		cli.BoolFlag{
+			Name:   "load-image",
+			Usage:  "load image from a file instead of building it",
+			EnvVar: "PLUGIN_LOAD_IMAGE",
+		},
+		cli.StringFlag{
+			Name:   "loaded-image-tag",
+			Usage:  "the tag of the loaded image",
+			EnvVar: "PLUGIN_LOAD_IMAGE_TAG",
```
oneacik, line 249: is it the tag of image loaded from file or is it tag not related to these two?
If it is unrelated then maybe it shouldn't be around there?

cmd/drone-docker/main.go
```diff
@@ -247,7 +262,10 @@ func main() {
 
 func run(c *cli.Context) error {
 	plugin := docker.Plugin{
-		Dryrun:  c.Bool("dry-run"),
+		Dryrun:            c.Bool("dry-run"),
+		Loadimage:         c.Bool("load-image"),
+		Loadedimagetag:    c.String("loaded-image-tag"),
```
oneacik, line 267: It is a duplication of string... the same string as in line 247.
Isn't there a way to deduplicate?

docker.go
```diff
@@ -133,7 +137,12 @@ func (p Plugin) Exec() error {
 		cmds = append(cmds, commandPull(img))
 	}
 
-	cmds = append(cmds, commandBuild(p.Build)) // docker build
+    if p.Loadimage == false {
```
oneacik, line 140: could it be extracted to the named function/lambda? It would better explain what is happening int this piece of code.

docker.go
```diff
@@ -276,6 +285,15 @@ func commandBuild(build Build) *exec.Cmd {
 	return exec.Command(dockerExe, args...)
 }
 
+func commandLoad(location string) *exec.Cmd {
```
oneacik, line 288: "loadDockerImageFromFile" name would be more suiting for this function, wouldn't it?

cmd/drone-docker/main.go
allgreed: Yup, they're all duplicated - however that's the plugin convention + it's contained to a single file

docker.go
allgreed: The convention seams to be "command" + actual Docker command, so I'm following that. Actually the name you've suggested implies sideffects, whereas the function deterministically produces a struct - it's a command pattern implementation

cmd/drone-docker/main.go
allgreed: So... the image yyyy... image [I'm referring to the .tar.gz that can be loaded] has a tag embedded in it, and I don't know how to extract that tag during plugin execution, so right now it needs to be explicitly passed [so that the plugin can operate on the loaded image as if it was builded in the default execution flow]. Does that answer your concern or did I misread something? I'm not completely sure I understand what you mean.

docker.go
allgreed: hmmm... I'd personally rather leave it as is, but there is definitely some point in getting the notion of producing image out there. Anyone has strong opinions on this?

allgreed:
Just a thought: I could also unzip the image and parse the manifest to read the embedded tag - seams like an overkill, but worth mentioning anyhow

allgreed:
@sbruder [since I have the rest of the people currently giving +1 on the speed dial] this problem can be solved by using `podman` with `--storage-driver=vfs` inside the container. Found out about it just know, will write a blog post soon, but this should get you started already if you're having this problem.

tboerger:
I would like to hear the opinion of @bradrydzewski on that.

allgreed:
@tboerger thank you for reaching out!
As of now I've created this utility (that looks *almost* like a proper plugin): https://github.com/allgreed/drone-load-and-store/

Usage example (I've tried to mimic this plugin's API): 
```
  image: allgreed/drone-load-and-store
  settings:
    archive: docker-image.tar.gz
    repo: username/repository
    username:
      from_secret: docker_username
    password:
      from_secret: docker_password
```

It's tested and works on my project. Also: having this as a separate plugin allowed me to solve the `load_image_tag` - no more of that!

So, from this point I'd be happy to either:
- try to carry this patch, IMO it'd be insanely cool if this problem had not existed in the first place and just having a go-to Docker plugin for "build-push" flow as well as "load-push" one
- close this PR and do some docs work (would appreciate a suggestion where this knowledge could be put sensibly) if you think that might be a good idea
- close this PR and leave the matter as is - I think that along a blog post I'm gonna write soon*ish* there is enough googlable trace for future Drone + Nix users (or any other people interested in "load-push" flow) having the same problem

docker.go
```diff
@@ -61,11 +61,15 @@ type (
...
-		Build   Build  // Docker build configuration
-		Daemon  Daemon // Docker daemon configuration
-		Dryrun  bool   // Docker push is skipped
-		Cleanup bool   // Docker purge is enabled
+		Login   Login         // Docker login configuration
+		Build   Build         // Docker build configuration
+		Daemon  Daemon        // Docker daemon configuration
+		Dryrun  bool          // Docker push is skipped
+		Cleanup bool          // Docker purge is enabled
+		Loadimage bool        // Docker load image is enabled
```
techknowlogick, line 69: Could you ensure PascalCase on these new options? (eg, `LoadImage`, etc..)

tboerger:
As you already created a plugin for this use case it would make sense to close this PR and submit the plugin with some docs to the plugin index.

allgreed:
@tboerger done, thank you for the suggesiton: https://github.com/drone/drone-plugin-index/pull/291

PR #296, opened 2020-08-05
Added env-file with godotenv to all non-base implementations of plugin

robertstettner:
I found that this didn't work for the Drone ECR plugin, as it was missing the `godotenv` getting environment variables from an env-file.


robertstettner:
Bump for merge @tboerger @bradrydzewski :)

Issue #286, opened 2020-06-08
Date label should be applied last


bjj:
I'm using the docker plugin with `cache_from` to make a build environment. It never changes, so literally half of my build time for that step is applying labels and then pushing those layers back to the registry. It would be nice if the only variable layer were applied last so all of the rest could be cached.

(I'm not sure why the labeling is so slow in general)

gtg3vv:
I'm also using cache_from for a rarely changed image, except that I would prefer to be able to turn off this label entirely to avoid unnecessary cache busting.

bradrydzewski:
sounds like a good request.  If someone can send a PR, we can take a look and merge quickly.

gtg3vv:
Done: #297

PR #297, opened 2020-08-12
#286 Add ability to exclude default build labels

gtg3vv:
* New no-default-labels flag to avoid cache busting via build-date label
* Reordered default label so build-date is applied last


gtg3vv:
Upon reviewing my tests, it looks like docker labels are sorted before being applied and the original request in the linked pr is not achievable. I don't think there is harm in leaving the line order as is, but I'm happy to remove that commit as well.

mj520:
There are four very good options for no-default-labels flag

tboerger:
As the current labels are deprecated I tend to drop them entirely and maybe add a flag to enable the replacement for the deprecated labels defined at https://github.com/opencontainers/image-spec/blob/master/annotations.md

PR #302, opened 2020-10-15
Add starlark build

donny-dont:
Use a starlark build.


bradrydzewski:
hey, I don't plan to adopt starlark for this repository.  The docker and bash plugins will follow a separate set of conventions from the other plugins.

donny-dont:
Closing in favor of #303

PR #303, opened 2020-10-15
Fix Windows build

donny-dont:
The Windows docker builds should only be triggered when there's a push to master or a tag. Also add proper tagging to the Windows builds.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -55,11 +55,9 @@ server:
     from_secret: windows_username
 
 steps:
-- name: build
+- name: build-push
   commands:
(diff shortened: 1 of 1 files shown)


tboerger:
I would prefer a starlark configuration similar to the other plugins instead of a bloated yml configuration.

donny-dont:
@tboerger I started a starlark one and @bradrydzewski said no so here we are 😄

tboerger:
@bradrydzewski why? this would improve the build definition and would be done in the same way as the other plugins.

PR #308, opened 2020-11-24
Add insecure mirror registry support

Nobuhira-Y:
Added an insecure_mirror option to allow mirror registry to be set as an insecure registry


tboerger:
The option is name insecure-registry, so why do you name it mirror now?

Nobuhira-Y:
insecure-registry is supposed to set up the Registry as insecure. When I want to set a mirror as insecure, I think insecure-registry is not sufficient
I don't know why insecure-registry is only for Registry.
But I consider it unusual to set a mirror as insecure, so I add a new option.
Should I allow insecure-registry to set both Registry and MirrorRegistry as insecure?

eliandoran:
@y-nobu , I would suggest we would use the more generic version described here: https://github.com/drone-plugins/drone-docker/pull/270

Instead of marking the mirror as insecure, simply add the repository to the list of insecure registries. This allows to use it in use cases other than the mirror.

@tboerger , by "mirror" I believe the OP wanted to allow the mirror repository to be insecure. The mirror repository being defined by this flag:

 > mirror
 > use a mirror registry instead of pulling images directly from the central Hub

hasafox:
So what option name can I use when I pull base image from insecure registry `A` and push product image to insecure registry `B`? Is `insecure_registry`, `insecure_mirror` or `mirror`?  I can't build my image using this plugin now.

eliandoran:
> So what option name can I use when I pull base image from insecure registry `A` and push product image to insecure registry `B`? Is `insecure_registry`, `insecure_mirror` or `mirror`? I can't build my image using this plugin now.

@hasafox , I guess you can use #270 , as per my comment here: https://github.com/drone-plugins/drone-docker/pull/270#issuecomment-1094273835

I think you have a similar use case, you'd probably have to add both A and B registries to the `insecure_registry` option. Unfortunately, since neither #308, nor #270 have been merged, you can only do this if you fork your own `drone-docker` repository and merge one of the PRs.

PR #309, opened 2020-11-25
add iptables option to the daemon

sherry-ummen:


cmd/drone-docker/main.go
```diff
@@ -103,6 +103,11 @@ func main() {
 			Usage:  "docker daemon executes in debug mode",
 			EnvVar: "PLUGIN_DEBUG,DOCKER_LAUNCH_DEBUG",
 		},
+		cli.BoolFlag{
```
ashwilliams1, line 106: change to `cli.BoolTFlag` so that it defaults to true

cmd/drone-docker/main.go
```diff
@@ -285,6 +290,7 @@ func run(c *cli.Context) error {
 			Disabled:      c.Bool("daemon.off"),
 			IPv6:          c.Bool("daemon.ipv6"),
 			Debug:         c.Bool("daemon.debug"),
+			IPTables:      c.Bool("daemon.iptables"),
```
ashwilliams1, line 293: change to `c.BoolT` flag so that it defaults to true

cmd/drone-docker/main.go
```diff
@@ -103,6 +103,11 @@ func main() {
 			Usage:  "docker daemon executes in debug mode",
 			EnvVar: "PLUGIN_DEBUG,DOCKER_LAUNCH_DEBUG",
 		},
+		cli.BoolFlag{
+			Name:   "daemon.iptables",
+			Usage:  "docker daemon enable addition of iptables rules",
+			EnvVar: "PLUGIN_IPTABLES",
```
ashwilliams1, line 109: change to `PLUGIN_IPTABLES_OFF`

docker.go
```diff
@@ -370,6 +371,9 @@ func commandDaemon(daemon Daemon) *exec.Cmd {
 	if daemon.Experimental {
 		args = append(args, "--experimental")
 	}
+	if daemon.IPTables {
```
ashwilliams1, line 374: need to verify the flag, but if `iptables_off: true` then we should disable iptables, which would otherwise be enabled by default

```
if daemon.IPTablesOff {
  args = append(args, "--iptables=false")
}
```

cmd/drone-docker/main.go
sherry-ummen: done

cmd/drone-docker/main.go
sherry-ummen: done

cmd/drone-docker/main.go
sherry-ummen: done

docker.go
sherry-ummen: done

sherry-ummen:
builds are failing not due to these changes . but due to the rate limit issue. if someone could still review and approve this

tboerger:
Why the name like iptables off? The flag could be true by default and the flag only gets appended if it's not true.

cmd/drone-docker/main.go
```diff
@@ -103,6 +103,11 @@ func main() {
 			Usage:  "docker daemon executes in debug mode",
 			EnvVar: "PLUGIN_DEBUG,DOCKER_LAUNCH_DEBUG",
 		},
+		cli.BoolTFlag{
```
bradrydzewski, line 106: I think this should be `Bool` instead of `BoolT`.  The `BoolT` parameter defaults to `true` if unset, however, in this case I think we want to disable to false and only disable IP tables if the user explicitly sets this value to `true`

cmd/drone-docker/main.go
```diff
@@ -103,6 +103,11 @@ func main() {
...
+			Name:   "daemon.iptables",
+			Usage:  "docker daemon enable addition of iptables rules",
```
bradrydzewski, line 108: also consider changing the description to make it clear that this flag disables iptable rules, instead of enabling

cmd/drone-docker/main.go
bradrydzewski: or as @tboerger mentioned, you can use `BoolT` to make it true by default, but then below, you need to change from `c.Bool` to `c.BoolT`

tphoney:
Have you had a look at implementing @bradrydzewski 's suggested changes and rebasing of master.

andriihrachov:
Any updates on this? highly needed

PR #312, opened 2021-01-17
add custom seccomp profile

xoxys:
This PR adds the default seccomp profile for docker 19.x and replace default action `SCMP_ACT_ERRNO` by `SCMP_ACT_TRACE`. This way Alpine Edge/3.13 ARM builds are working again using Docker 19.x. For details see: https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.13.0#time64_requirements

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -341,7 +341,10 @@ func commandDaemon(daemon Daemon) *exec.Cmd {
 	args := []string{
 		"--data-root", daemon.StoragePath,
 		"--host=unix:///var/run/docker.sock",
-		"--seccomp-profile=/etc/docker/default.json",
+	}
+
+	if _, err := os.Stat("/etc/docker/default.json"); err == nil {
+		args = append(args, "--seccomp-profile=/etc/docker/default.json")
 	}
 
 	if daemon.StorageDriver != "" {
--- a/docker.go
+++ b/docker.go
@@ -341,6 +341,7 @@ func commandDaemon(daemon Daemon) *exec.Cmd {
 	args := []string{
 		"--data-root", daemon.StoragePath,
 		"--host=unix:///var/run/docker.sock",
+		"--seccomp-profile=/etc/docker/default.json",
 	}
 
 	if daemon.StorageDriver != "" {
--- a/docker/docker/Dockerfile.linux.amd64
+++ b/docker/docker/Dockerfile.linux.amd64
@@ -2,5 +2,14 @@ FROM docker:19.03.8-dind
 
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 
+RUN apk --update add --virtual .build-deps curl && \
+    mkdir -p /etc/docker/ && \
+    curl -SsL -o /etc/docker/default.json https://raw.githubusercontent.com/moby/moby/19.03/profiles/seccomp/default.json && \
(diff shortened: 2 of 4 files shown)


xoxys:
Related issue https://github.com/docker-library/docker/issues/260

ashwilliams1:
I suggest we should upgrade the images to use `docker:20.10-dind` to solve this issue, as opposed to patching the seccomf file.  Upgrading to a newer version of Docker seems like a more permanent solution.

xoxys:
There is no docker:20-dind for arm32 see linked issue...

xoxys:
This one is linked to the one I posted as reference https://github.com/docker-library/docker/issues/286

ashwilliams1:
I see.  In that case I suggest we only patch the seccomp file for arm, since the arm64 and amd64 images are currently working without issue.

xoxys:
And how? Just remove the `sed` command? As the file needs to be there because of `--seccomp-profile=/etc/docker/default.json`.

ashwilliams1:
perhaps I am misunderstanding this change, but we are not currently experiencing any issues with arm64 or amd64 machines so I think it is best not to change the dockerfiles for this architecture.  If you are experiencing issues with arm32 machines we should only update the arm32 dockerfile, as a temporary workaround, which we can revert when `docker:dind-20` is available for arm32.  Or perhaps downgrade your arm32 pipelines to use `plugins/docker:1.18` and then we wait until Docker has an official solution.

ashwilliams1:
> As the file needs to be there because of --seccomp-profile=/etc/docker/default.json

It should be trivial to only load a seccomp profile if the architecture is arm:

```
if os.GOARCH == "arm" {
  args = append(args, "--seccomp-profile=/etc/docker/default.json")
}
```

xoxys:
And still no idea how this should work... To load a custom seccomp profile we need to change the source code of the plugin go binary (see `docker.go`). As the source code (and resulting binary) is the same for every docker image afaik we can't just change the dockerfile for a single docker image.

xoxys:
Alright if that's the way you want to handle it, I can add it... Personally I don't use this plugin anymore, so I don't really care about the way to implement this. You just need to tell me how you want it

xoxys:
> edit: this may be even better:
> 
> ```
> if _, err := os.Stat("/etc/docker/default.json"); err == nil {
>   args = append(args, "--seccomp-profile=/etc/docker/default.json")
> }
> ```

Done.

tboerger:
IMHO it's better to keep the different architectures on the same style. I can't see anything against using the upstream seccomp profile. @bradrydzewski please write down your opinion. The drone cloud isn't useable for months for multi arch docker images based on alpine linux.

bradrydzewski:
I have always felt strongly that we should not fix something that isn't broken.  The arm64 and amd64 images are working fine and I would hate to see an unnecessary change break something due to an edge case that we did not anticipate.  I think we should implement this as a temporary workaround for arm32 only, and then when docker-in-docker is available for 20.x we can remove the temporary workaround.

xoxys:
Anything missing getting this merged? CI is failing due to docker pull rate limit.

xoxys:
@ashwilliams1 @bradrydzewski any blocker?

PR #319, opened 2021-04-11
feature: added support the autolabel env variable for acr/ecr/gcr/heroku

YouSysAdmin:
In plugin versions acr/ecr/gcr/heroku "auto_label: false" does not work.
Added support for this parameter 

```yaml
  - name: build-and-push
    image: plugins/ecr:19.03.8
    settings:
      ...
      auto_label: false
```
Before:
```sh
/usr/local/bin/docker build --rm=true -f Dockerfile -t ..b7c17b2.. . --pull=true --label org.label-schema.schema-version=1.0 --label org.label-schema.build-date=2021-04-11T19:18:47Z --label org.label-schema.vcs-ref=..b7c17b2.. --label org.label-schema.vcs-url=https://github.com/user/repo.git
```
After:
```sh
/usr/local/bin/docker build --rm=true -f Dockerfile -t ..b7c17b2.. . --pull=true
```

### DIFF
--- a/cmd/drone-acr/main.go
+++ b/cmd/drone-acr/main.go
@@ -16,10 +16,11 @@ func main() {
 	}
 
 	var (
-		repo     = getenv("PLUGIN_REPO")
-		registry = getenv("PLUGIN_REGISTRY")
-		username = getenv("SERVICE_PRINCIPAL_CLIENT_ID")
-		password = getenv("SERVICE_PRINCIPAL_CLIENT_SECRET")
+		repo      = getenv("PLUGIN_REPO")
+		registry  = getenv("PLUGIN_REGISTRY")
+		username  = getenv("SERVICE_PRINCIPAL_CLIENT_ID")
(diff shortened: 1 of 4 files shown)


bradrydzewski:
these changes are not required.  All environment variables are automatically passed to the child process [1].  

> Env specifies the environment of the process.
> Each entry is of the form "key=value".
> If Env is nil, the new process uses the current process's
> environment.

[1] https://golang.org/pkg/os/exec/#Cmd

YouSysAdmin:
> These changes are not required. All environment variables are automatically passed to the child process [1].

Sorry, the tired brain played a cruel joke :)
I know that it should work like that, but it didn't work for me, and I found my mistake - the plugin version!

I took the tag 19.03.8, assuming that this is the latest version and did not pay attention to the date, this version was released 5 months ago and the "AUTO_LABEL" variable was added 3 months ago :)

In the master branch contains the version without tag :(

PR #321, opened 2021-06-08
Creates support for an additional registry

BitProcessor:
Usecase: 
* Main registry credentials used for pushing the resulting container towards registry 1 (example: ECR)
* Additional credentials used for pulling base images (FROM: ) from another registry 2 that also needs authentication (example: pulling images from a nexus-proxy)

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -219,19 +219,19 @@ func main() {
 			EnvVar: "PLUGIN_PASSWORD,DOCKER_PASSWORD",
 		},
 		cli.StringFlag{
-			Name:   "docker.registry.alternative",
-			Usage:  "docker alternative registry",
-			EnvVar: "PLUGIN_REGISTRY_ALTERNATIVE,DOCKER_REGISTRY_ALTERNATIVE",
+			Name:   "docker.registry.additional",
+			Usage:  "docker additional registry",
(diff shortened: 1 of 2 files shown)


bradrydzewski:
you can use the `config` attribute for this, which loads a `.docker/config.json` file

```
settings:
  config: |
    {
      "auths": ...
    }
```

BitProcessor:
> you can use the `config` attribute for this, which loads a `.docker/config.json` file. This file can include the credentials for one or more regsitries.
> 
> ```
> settings:
>   config: |
>     {
>       "auths": ...
>     }
> ```

Would that work together with ECR ? 
My use-case is: 
* Registry 1 contains the base images and needs authentication (simple username/password)
* Registry 2 == ECR and the destination for the resulting container image

bradrydzewski:
Yep!  use the ecr plugin so that you can authenticate with Registry 2, and then provide the credentials for Registry 1 using the config parameter.

BitProcessor:
@bradrydzewski magic 🎉 
It works like a charm.

Final question: I suppose this method doesn't support any form of using Drone secrets ?

```
{
	"auths": {
		"my-repo.example.com": {
			"auth": "somesecret"
		}
	}
}
```

bradrydzewski:
> I suppose this method doesn't support any form of using Drone secrets ?

in this case, you would store the entire contents of config as a secret, for example:

```
settings:
  config:
    from_secret: ...
```

PR #322, opened 2021-06-27
Update seccomp to 20.10 docker

techknowlogick:
Alpine 3.14 requires update seccomp profile

PR #306, opened 2020-11-18
Upgrade docker-dind to 19.03.13

ademariag:
Docker based plugins are currently broken. This PR upgrades the base image to use 19.03.13

https://discourse.drone.io/t/broken-builds-with-docker-19-03-13/8373/7

PR #327, opened 2021-07-13
Update Docker dind to 20.10.7-dind

### LINKED ISSUE #326: Upgrade to Docker 20.10.05
Hi 👋

Due to the fact that the Drone Docker plugin still uses Docker v19.03.8:

```#!sh
latest: Pulling from plugins/docker
3s
2	Digest: sha256:1088e4e26b87ac4c24948ec2d2a738cea1f9e4e9a7b81eef3b77eae41d57e978
3s
3	Status: Image is up to date for plugins/docker:latest
3s
4	+ /usr/local/bin/dockerd --data-root /var/lib/docker --host=unix:///var/run/docker.sock
4s
5	Registry credentials or Docker config not provided. Guest mode enabled.
6s
6	+ /usr/local/bin/docker version
6s
7	Client: Docker Engine - Community
6s
8	 Version:           19.03.8
...
```

One will run into the following error when using `make` and a `Makefile` in conjunctionn with an Alpine image 3.14+:

```#!sh
Step 9/21 : RUN make TAG=$TAG BUILD=$BUILD build
83s
169	 ---> Running in 5371b7f93f10
83s
170	make: rm: Operation not permitted
84s
171	make: *** [Makefile:33: clean] Error 127
```

This was filed as a [bug report](https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2) upstream with Alpine which linked to [this set of fixes and work-arounds](https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2).

The simplest fix is to upgrade the  Docker version used here.

Can we do that please?

prologic:
Fixes #326

I believe this resolves the issue I described in #326

I'm not sure how to test this myself, without just building my own plugins/docker
which I haven't done bfore...

### DIFF
--- a/docker/docker/Dockerfile.linux.arm
+++ b/docker/docker/Dockerfile.linux.arm
@@ -1,4 +1,4 @@
-FROM arm32v6/docker:19.03.8-dind
+FROM arm32v6/docker:20.10.7-dind
 
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 
--- a/docker/docker/Dockerfile.linux.amd64
+++ b/docker/docker/Dockerfile.linux.amd64
@@ -1,4 +1,4 @@
-FROM docker:19.03.8-dind
+FROM docker:20.10.7-dind
 
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 
--- a/docker/docker/Dockerfile.linux.arm64
+++ b/docker/docker/Dockerfile.linux.arm64
@@ -1,4 +1,4 @@
-FROM arm64v8/docker:19.03.8-dind
+FROM arm64v8/docker:20.10.7-dind
 
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 


bradrydzewski:
There is a problem with the official Docker-in-Docker images that prevents upgrading at the moment. See https://github.com/drone-plugins/drone-docker/pull/306#issuecomment-754305557

CirnoT:
Either Docker version must be upgraded or https://github.com/drone-plugins/drone-docker/pull/322 must be merged again (can't see why it would cause mentioned issues, possibly these were stray issues unrelated to changes made or possibly users are using cgroupsv2) or we will block using Alpine 3.14 images forever.

In addition, the old version of Docker prevents usage of cgroupsv2.

techknowlogick:
@CirnoT I've built and pushed my fork here with the linked PR: https://hub.docker.com/r/techknowlogick/drone-docker for arm64&amd64 in case you need it.

prologic:
> There is a problem with the official Docker-in-Docker images that prevents upgrading at the moment. See [#306 (comment)](https://github.com/drone-plugins/drone-docker/pull/306#issuecomment-754305557)

Is it worth holding back for one platform? THis is CI we're talking about here, I'd imagine most CI runs are on amd64 and presumably on Linux anyway. Without the update a lot of folks's workflows that depend on Make and use Alpine are broken.

prologic:
So I worked out how to build and use this in my environment, but there are _problems_ with it. See attached log.

[logs_yarnsocial_yarn_1_3.log](https://github.com/drone-plugins/drone-docker/files/6812848/logs_yarnsocial_yarn_1_3.log)

CirnoT:
> Is it worth holding back for one platform? THis is CI we're talking about here, I'd imagine most CI runs are on amd64 and presumably on Linux anyway.

ARM (more specifically however ARM64) is gaining more traction every day, abandoning it here would bring more bad than the Alpine 3.14 issue.
While Drone definitely needs to support ARM32 v6/v7, is it worth keeping DID support for this architecture still @bradrydzewski? I imagine nowadays most CI/CD aimed at this architecture would be native code for embedded applications. I suppose Docker here is of similar opinion as they clearly don't consider this platform of importance enough to fix the issue for months. Frankly I suspect they'd drop it entirely rather than fixing it.

prologic:
Regardless of the support of other arches  (_I agree they should continue to be supproted_), I can't even get the Docker daemon started when upgrading the plugin to use `20.10.7-dind`. There are numerous problems with the startup, mostly around "Operation not permitted" and IPTables rubbish. I'm _actually_ not sure how to resolve any of these at the moment...

prologic:
Ahhh I found what I needed

[DRONE_RUNNER_PRIVILEGED_IMAGES](https://docs.drone.io/runner/docker/configuration/reference/drone-runner-privileged-images/)

techknowlogick:
@prologic a note, if you use the autoscaler the env var you'll need to set for the autoscaler itself is `DRONE_RUNNER_PRIVILEGED`

prologic:
I got this working successfully. So this PR _at least_ is functional. One thing to note for anyone running Drone and Drone Runner(s) in Docker Swarm is to ensure your runners have access to the default `bridge` network if you intend to go out over the Internet and back again like I do for _some_ things (_stupid I know, I should fix this one day_) -- If you don't get end up with weird obscure errors like:

```
dial tcp: lookup ... on 127.0.0.11:53: read udp 127.0.0.1:xxxxx->127.0.0.11:53: read: connection refused
```

CirnoT:
@bradrydzewski Debian 11 has been released along with new systemd and unified cgroupsv2. Docker 19 does not support cgroupsv2 and DID will always fail on such setup. Workaround of `systemd.unified_cgroup_hierarchy=0` is unsupported by systemd and heavily discouraged by Debian maintainers.

Is this repository still being maintained or are we to assume that Drone has no intent to move forward with technology and replacement should be found?

docker/docker/Dockerfile.linux.arm
```diff
@@ -1,4 +1,4 @@
-FROM arm32v6/docker:19.03.8-dind
+FROM arm32v6/docker:20.10.7-dind
```
CirnoT, line 1: Reverting back to 19.03.8 for arm32 should allow us to move forward with 20.10 on rest of platforms while keeping (on life support, as it seems Docker very much does not care about that platform anymore) arm32 usable

https://github.com/docker-library/docker/issues/260
https://github.com/drone-plugins/drone-docker/pull/306

bradrydzewski:
for `docker/docker/Dockerfile.linux.arm` we need to use `FROM arm32v6/docker:19.03.8-dind` since docker is no longer publishing arm32 images.  So if we revert changes to `docker/docker/Dockerfile.linux.arm` we can merge this pull request.  thanks!

johnmarksilly:
Bump +1 My team is currently stuck manually building our prod images locally due to this CI issue :(

eoinmcafee00:
Closing as no longer required.
Updated in latest release https://github.com/drone-plugins/drone-docker/releases/tag/v20.10.9

PR #332, opened 2021-09-16
Secrets enable in Docker plugin with support for netrc file

wszychta:
### PR purpose
This PR is providing options to:
- Use secrets in docker plugin
- Pass `~/.netrc` file as secret to the build

Also when:
- if any `secrets` are provided
- `git_netrc_pass` is set to `true` 

then environment variable `DOCKER_BUILDKIT` is set to `1`. If nothing is provided then plugin willl not create/change that variable.

### Warning
Because of [urfave/cli](https://github.com/urfave/cli) library is not able to Slice Flags with different sign than `,` then we have introduced another variable called `secret_separator`. The content of this variable will be replaced with `,` on docker build command creation.

### Example usage
`.drone.yml`:
```
- name: test Docker
  image: plugins/docker
  settings:
    dry_run: true
    repo: foo/bar
    tags: latest
    dockerfile: Dockerfile
    git_netrc_pass: true
    secret_separator: ":"
    secrets: 
      - id=test-secret:src=test-file.txt
      - id=test-secret2:src=test-file2.txt
```
`Dockerfile`:
```
# syntax=docker/dockerfile:1.2
ARG IMAGE_VERSION=latest
FROM ubuntu:${IMAGE_VERSION}
RUN --mount=type=secret,id=test-secret,dst=/root/test-secret.txt cat /root/test-secret.txt
RUN --mount=type=secret,id=test-secret2,dst=/root/test-secret2.txt cat /root/test-secret2.txt
RUN apt update; apt install -y git
RUN --mount=type=secret,id=git-netrc,dst=/root/.netrc git clone https://github.com/PRIVATE/REPO/PATH.git
```

### Other work to do
This PR requires update to drone [docker plugin documentation](http://plugins.drone.io/drone-plugins/drone-docker/)


bradrydzewski:
👋  Hey there, while I appreciate the pull request, I recommend reading this thread:
https://discourse.drone.io/t/drone-netrc-username-documentation-usage/4485/2

The netrc file is considered _Internal use only_ and is subject to change. We are working to narrow the use of the netrc file and if we can find a suitable alternative, we may consider removing in the future.  If you want to inject git credentials into this plugin it should be done using [secrets](https://docs.drone.io/secret/repository/).

In terms of using docker secrets, it might make more sense to model this behavior after [`build_args_from_env`](https://discourse.drone.io/t/how-can-i-pass-secrets-as-build-args-to-plugins-docker-drone-0-8/824/2) as opposed to reading values from a file.  I recommend working with the maintainers to design the yaml syntax before writing additional code.

bradrydzewski (changes requested):
please remove netrc functionality from this pull request.  we should scope this pull request around the use of the `--secrets` flag.

wszychta:
@bradrydzewski I have modified code as you suggested. I have also modified my Pull Request description with new changes. Right now this Pull Request doesn't create `netrc` file automatically.

bkk-bcd:
@bradrydzewski this looks like something we could use today, can you take another look to help get this merged in?

bkk-bcd:
Looks like the build container is using go `1.11` while go mod requires go `1.13`.

Go `1.13` has `os.UserHomeDir()`

Not until go `1.16` does `os.WriteFile` exists.

bradrydzewski:
I think a more thorough design review is going to be required and I want to be up front that this will take some time.  We have a strong commitment to backward compatibility which means once this feature is merged it cannot easily be changed, so we need to make sure we are happy with the design and implementation.

wszychta:
@bradrydzewski ok, I will wait until you will review this PR in next sprint.
Also I would suggest to change go version on vet to newer than `1.11`.
Locally I'm working on go `1.16` and I didn't have any problems with docker and ECR plugins, but I was unable to check other plugins.

alexef:
Looks like this could be closed now that #359 was merged.

wszychta:
@alexef I'm not working for the company which was requiring this feature. I will contact somebody from them and ask for comment here.

bmmij:
We will, test out the release https://github.com/drone-plugins/drone-docker/releases/tag/v20.12.0, thats is refered in here.

PR #346, opened 2021-12-03
Support for windows images for tags

shubham149:


tphoney (changes requested):
@shubham149 can you squash your commits and remove any commented out code as it is distracting.

Due to the importance of this plugin, it would be good to get @bradrydzewski to look over it too.

bradrydzewski:
a few thoughts ...

First, I recommend using powershell scripts to build and publish the image, similar to what we do here:
https://github.com/drone-runners/drone-runner-docker/blob/master/.drone.yml#L167:L190
https://github.com/drone-runners/drone-runner-docker/blob/master/scripts/windows/tag.ps1
https://github.com/drone-runners/drone-runner-docker/blob/master/scripts/windows/latest.ps1

The above powershell scripts have more robust tagging strategies and I personally find it a little cleaner having the powershell in separate scripts, as opposed to in the yaml.

Second, given the lack of support for docker-in-docker, this plugin does not work very well on windows.  It requires mounting the host machine named pipe on windows.  Do we want to continue expanding support for this plugin on windows given it is subpar?  or do we want to have a windows-optimized alternative?

shubham149:
As discussed, we will explore binding docker daemon socket of the vm to the docker container and use that for building images of container on windows since privileged mode is not enabled.

tphoney (approved):

PR #347, opened 2022-01-05
(feat) publish docker data to create drone card

eoinmcafee00:
On merge - update page setting to point to master & not branch


docker.go
```diff
@@ -204,6 +297,17 @@ func commandPull(repo string) *exec.Cmd {
...
 
+func commandInspect(build Build, tag string) *exec.Cmd {
+	target := fmt.Sprintf("%s:%s", build.Name, tag)
+	args := []string{
+		"docker inspect",
+	}
+	args = append(args, target)
+	args = append(args, "> /tmp/output.json")
+
+	return exec.Command(bashExe, "-c", strings.Join(args, " "))
```
shubham149, line 308: bash won't be present in windows image. Also, /tmp directory would be present in windows.

shubham149: Use this command instead:
`
exec.Command("docker", "inspect")
`
and redirect stdout and stderr to /tmp/output.json file

eoinmcafee00: @shubham149 - I missed your updates where you've added in windows support, which is the reason this was missed. Will look into it

bradrydzewski: instead of using bash and writing to a file, you can read directly from `exec.Cmd.Stdout`

eoinmcafee00: yeah this has been cleaned up.

bradrydzewski (approved):

PR #349, opened 2022-01-06
Testing windows pipeline

shubham149:


bradrydzewski:
for security reasons, we cannot allow pull requests to execute on our windows machines, otherwise a bad actor could submit a pull request that installs malware on the machine (which would impact subsequent pipelines)

PR #357, opened 2022-02-14
Add support for ECR Public registry

tlinhart:
Adds support for the [ECR Public](https://aws.amazon.com/blogs/aws/amazon-ecr-public-a-new-public-container-registry/) registry. There is a fork (https://github.com/cstyan/drone-docker) which can already be used for the purpose. However, it produces Docker image only for x86-64 architecture and it might be a good idea to have it under one roof with other registries.


tphoney:
Hi @tlinhart thanks for the PR, to start with. 
Just to confirm your intentions, this is to enable pushing the drone-docker plugin to ECR. 
If you need images mirrored you could use https://pypi.org/project/ecr-mirror/
We dont have a presence for drone in ECR as yet, and it would be a major undertaking for us to add another registry.

tlinhart:
Hi @tphoney, this PR doesn't add support for pushing the drone-docker image itself to ECR. It adds support for building and pushing images to ECR Public registry in the pipeline step. Just like for other already supported registries (ECR, GCR etc.)

The main reason for this PR is that Loki repository (by Grafana) uses Drone and ECR Public as the target registry for their Promtail Lambda tool. I opened a PR for adding ARM64 support (https://github.com/grafana/loki/pull/5354) but there was a problem in that there was no ARM64 runner image (i.e. drone-docker plugin) for the ECR Public. So I thought it would be a generally good idea to add the support for another possibly widely used registry to the upstream.

masterkain:
I too need to push images to public ECR -- this caught me a bit off guard since kaniko plugins were able to without particular config and I did a lot of trial and errors before landing here.

niondir:
We are also looking into pushing into the public ECR registry ...

PR #362, opened 2022-04-27
Dry-run instead of skip, when auto_tag: true and no tag is matched

gzm0:
This safely allows to check larger parts of the build
earlier (e.g. during a pull request).

This is changes existing behavior, but this is IMHO better than
introducing yet another option. Mainly for the following reasons:

- This is the behavior I'd expect.
- The old behavior can be easily recreated using `when.ref`.

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -333,8 +333,8 @@ func run(c *cli.Context) error {
 			}
 			plugin.Build.Tags = tag
 		} else {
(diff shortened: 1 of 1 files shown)


bradrydzewski:
The concern I have is that secrets are disabled by default for pull requests which means registry credentials are not guaranteed to be available to execute a dry run (in cases where the Dockerfile base image is private).  This could result in a breaking change for pipelines that rely on the current behavior.

gzm0:
Totally valid concern indeed :-(. I'm closing this for now.

PR #364, opened 2022-05-21
Use Docker Buildx to build images

ste93cry:
Docker Buildx is a plugin that extends the `docker` command wtih additional features provided by [Moby BuildKit](https://github.com/moby/buildkit) builder toolkit. One of the features that most benefit from using Buildx is the multiarch build: without the plugin, to build and tag an image for more than one architecture you need to build each Docker image separately, push them to the registry and then create a manifest list with `docker manifest`. Instead, when Buildx is available, it's enough to build the image as usual, taking care of using the additional `--platform` argument. With this PR, I'm installing the plugin and I'm switching how the Docker images are built to always use it. For existing pipelines nothing should change as the plugin, for what I've seen, is fuilly backwards compatible. In a subsequent PR, if this one gets accepted, I plan to add the support for the aforementioned `platform` option


bradrydzewski:
Please see our response to a similar proposal:
https://github.com/drone/proposal/issues/5#issuecomment-878472674

There is also an existing buildx plugin that the community can leverage:
https://github.com/thegeeklab/drone-docker-buildx

We are certainly interested in an official buildx plugin, however, for reasons described in the proposal we would prefer this be launched as a separate plugin at this time.  If there is interest, we can work with the community to create a separate repository where work on a buildx plugin can begin.

ste93cry:
While I somehow understand your concerns about the stability of this plugin, I see a lot of cons in having a separate one: specifically, maintaining two plugins means that the changes to one have to be replicated in the other, and if their codebase start diverging it wil be even more difficult to keep them in sync. According to the official documentation, `buildx` is just a plugin that extends the standard `docker` command, so I don't see how the stability of this Docker plugin could be affected: the risk of introducing bugs is the same of other PRs

bradrydzewski:
as previously mentioned, we are happy to work with the community to create a separate repository where we can explore creating a buildx alternative to the traditional docker plugin.  Here is a new repository where such work can begin, without needing to be concerned with backward compatibility <https://github.com/drone-plugins/drone-buildx>

PR #366, opened 2022-06-01
Add support for docker --platform flag

mem:
Add a --platform flag so that the user can specify a value for the
corresponding flag passed to docker. This is useful in the context of
multi-arch builds and buildkit (DOCKER_BUILDKIT=1 in the environment).

Signed-off-by: Marcelo E. Magallon <marcelo.magallon@grafana.com>

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -269,6 +269,11 @@ func main() {
 			Usage:  "card path location to write to",
 			EnvVar: "DRONE_CARD_PATH",
 		},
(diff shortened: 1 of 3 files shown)


steven-mi:
@bradrydzewski Could you have a look?

tphoney:
Hi @mem i had to take your PR and rebase it to master #376 . Many thanks for this great change. 
I will update the documentation too on https://plugins.drone.io/plugins/docker as well doing a release.

PR #368, opened 2022-06-09
[Feature] support keep intermediate container for multi-stage build caching

zjc17:
fore using `--rm=true` makes trouble for multiple stage build. The `always remove intermediate container` behavior prevents people to speed up their building.

Here is the screenshot when --rm=true.
image

And it should be cached as it's built in jenkins

image

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -153,6 +153,11 @@ func main() {
 			Usage:  "quiet docker build",
 			EnvVar: "PLUGIN_QUIET",
 		},
+		cli.BoolFlag{
+			Name:   "ignore-rm",
+			Usage:  "remove intermediate container",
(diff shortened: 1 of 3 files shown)


zjc17:
I've tested it ...
image

tphoney (changes requested):
Hey @zjc17 i really like the idea of this feature, with one small proviso. 
There is already the purge command, does it overlap with what you suggest ?https://github.com/drone-plugins/drone-docker/pull/368/files#diff-60e9b6677b8d8d34db5104f9d30dc207928396046b5e06cc9f808508435775bcL232-L237

zjc17:
@tphoney cool! It works for me. Thx

tphoney:
glad i could help shout out to @bradrydzewski for the tip

PR #370, opened 2022-07-14
Add Power(ppc64le) support

sumangalm-1:
Add Power(ppc64le) support

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -670,7 +670,7 @@ steps:
     dockerfile: docker/docker/Dockerfile.linux.ppc64le
     password:
       from_secret: docker_password
-    repo: sbm13/drone
+    repo: plugins/docker
     username:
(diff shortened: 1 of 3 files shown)


jimsheldon:
Unfortunately this will not work because there is not a ppc64le docker runner available to execute this pipeline.

If we were using something like [GoReleaser](https://goreleaser.com/) and/or [Docker Buildx](https://docs.docker.com/buildx/working-with-buildx/) we wouldn't need native ppc64le hardware.

I will investigate how much work it would be to use one of those projects with this repo.

eoinmcafee00:
Hey @sumangalm-1 

Closing this PR for now due to @jimsheldon comment above. 

Thank's again for the submission.

uPagge:
Do I understand correctly that with the help of your plugin, I cannot build an image for apple m1?

PR #373, opened 2022-09-14
Use full path to docker when creating card

donny-dont:
Windows docker builds failed to create the card using `docker inspect` because docker isn't on the path. Instead use `dockerExec` from `daemon.go` which has the full path to the executable.

### DIFF
--- a/card.go
+++ b/card.go
@@ -18,7 +18,7 @@ import (
 )
 
 func (p Plugin) writeCard() error {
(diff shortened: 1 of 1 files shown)


tphoney (approved):

tphoney:
thanks for the fix, i still have the tabs open from when you mentioned it. You got their first many thanks
🥂

PR #374, opened 2022-10-14
Support auto_tag with dry-run when no tag will be created

gzm0:
This safely allows to check larger parts of the build earlier (e.g. during a pull request).

An original version did not introduce an additional flag but changed the existing behavior. However it was rejected because it would break builds relying on non-PR secrets to pull the base images for the build.


gzm0:
Alternative to #362.

PR #380, opened 2022-12-19
Multiplatform build with Buildx

steven-mi:


luigibarbato:
This PR is still in progress? it would be really useful :)

steven-mi:
@luigibarbato We opened the PR draft to show others how to refactor this plugin to use it for multi-platform image creation. However, we do not plan to work on it continuously as it is not a high priority for us.

PR #381, opened 2023-01-17
(maint) move to harness.drone.io

tphoney:


.drone.yml
```diff
@@ -1,1185 +1,991 @@
...
-- linux-amd64-docker
+  - linux-amd64-docker
 
 ---
 kind: pipeline
 name: linux-arm64-heroku
+type: vm
+
+pool:
+  use: ubuntu
```
eoinmcafee00, line 756: Is this the correct pool?

tphoney: good catch, change incoming

eoinmcafee00 (approved):

PR #383, opened 2023-01-24
Allow gcr authentication with workload identity

dhpollack:
This allows the GCR plugin to use a [workload identity federation](https://cloud.google.com/iam/docs/workload-identity-federation) key.  This uses a special type of config file to generate a temporary oauth token for drone.  This is recommended by google over using service account keys.


rohit-gohri:
Hey @dhpollack Could you explain how you are generating the token for workload identity? Are you storing it in secrets

rohit-gohri:
I'm trying to recreate a setup similar to https://github.com/google-github-actions/auth in drone

dhpollack:
> Hey @dhpollack Could you explain how you are generating the token for workload identity? Are you storing it in secrets

To generate a token, you don't necessarily need workload identity.  You can use the gcloud command:

```
gcloud auth application-default print-access-token --quiet
```

to generate an actual workload identity config you can run the command:

```
export GCP_PROJECT_NUMBER=[your project number (not name)]
export POOL_ID=[name of workload identity pool]
export PROVIDER_ID=[name of your provider. related to where the request is coming from]
export SERVICE_ACCOUNT_EMAIL=[service account email]
gcloud iam workload-identity-pools create-cred-config \
  "projects/${GCP_PROJECT_NUMBER}/locations/global/workloadIdentityPools/${POOL_ID}/providers/${PROVIDER_ID}" \
  --service-account=${SERVICE_ACCOUNT} \
  --aws \
  --output-file="$HOME/.config/application_default_credentials.json"
```

dhpollack:
Setting up the pool itself is not entirely trivial.  We used terraform, but there was a lot of playing around with the settings:

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iam_workload_identity_pool
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iam_workload_identity_pool_provider

The official google instructions are also helpful to know what all the options are:

https://cloud.google.com/iam/docs/workload-identity-federation

dhpollack:
Looking at the docs that you pointed to, you need to create an `access_token`.  The service account json method is much easier but as stated in the documentation, it's not as secure.

alzabo:
Not sure if this will help get time from maintainers for this PR, but the patch works fine for me.

As @dhpollack writes in his initial comment, use of service account key export is not recommended by Google

dhpollack:
I found a bug with base64 encoded workload identity files, which I fixed.  I also found a bug in the drone-ecr plugin where the default value for the env vars was always set to false for booleans.

cmd/drone-gcr/main.go
```diff
@@ -61,12 +60,52 @@ func main() {
...
+	if err == nil {
+		token, err := credentials.TokenSource.Token()
+		if err == nil {
+			return token.AccessToken
+		}
+	}
+	return
+}
+
+func setUsernameAndPassword(user string, pass string, is_wi bool) (u string, p string) {
```
tphoney, line 84: golang uses camelcase or pascal case for variable names depending on context, it should not use snake case. 
https://gyulgrigoryan.medium.com/variables-in-golang-1bf492df03b0
so `is_wi` can you rename it to `workloadIdentitiy"

tphoney (changes requested):
Great fix for the bugged method in ecr 🥂 
there is a couple of nit picks around variable naming, nothing big.
Also could you squash your commits into either one commit with everything (we can do that at merge time), or 2 commits with the ecr fix in one, and the gcr work in the other. your choice.
thanks for the effort and expertise !!!

dhpollack:
@tphoney I do have one more question.  Does the documentation for this automatically get generated or do I need to make the changes somewhere?  I couldn't find where that was done.

tphoney (approved):

tphoney:
thanks @dhpollack your go code was fine, it was just to keep it in line with the other code. 

https://github.com/drone/drone-plugin-index/blob/main/plugins/gcr/content.yaml is the raw documentation for the the plugins website here https://plugins.drone.io/plugins/gcr

once again thanks for the change !

dhpollack:
@tphoney I created a PR for the docs.

drone/drone-plugin-index#414

PR #385, opened 2023-02-24
fix: assign a unigue build name based on commit SHA and image name

philanton:


Billdex:
Thanks for your work, but may be it's not enough. 
I have the same problem when I parallel build a repo with different parameters.
I think we can generate build name with DRONE_BUILD_NUMBER to fix it.

PR #386, opened 2023-03-22
Add support for cache backend

rutvijmehta-harness:
WIP

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -164,10 +164,11 @@ func main() {
 			Usage:  "images to consider as cache sources",
 			EnvVar: "PLUGIN_CACHE_FROM",
 		},
-		cli.StringFlag{
+		cli.GenericFlag{
 			Name:   "import-cache",
 			Usage:  "cache import location",
 			EnvVar: "PLUGIN_IMPORT_CACHE",
+			Value:  new(docker.CustomStringFlag),
 		},
 		cli.StringFlag{
 			Name:   "export-cache",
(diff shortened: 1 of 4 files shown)


cmd/drone-docker/main.go
```diff
@@ -324,6 +335,8 @@ func run(c *cli.Context) error {
 			Quiet:       c.Bool("quiet"),
 			Platform:    c.String("platform"),
 			SSHAgentKey: c.String("ssh-agent-key"),
+			ImportCache: decodeImportCache(c.String("import-cache")),
```
shubham149, line 338: Create a custom variable instead of StringFlag. https://github.com/drone-runners/drone-runner-aws/pull/237/files#diff-cf7fbb42c66fce9a98485449d21136aba758d0974fabd1c18a03dff5749dcc23R12

rutvijmehta-harness: Done

docker.go
```diff
@@ -65,6 +65,8 @@ type (
 		Platform    string   // Docker build platform
 		SSHAgentKey string   // Docker build ssh agent key
 		SSHKeyPath  string   // Docker build ssh key path
+		ImportCache []string // Docker buildx import cache
+		ExportCache string   // Docker buildx export cache
```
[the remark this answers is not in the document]
rutvijmehta-harness: From [docs](https://docs.docker.com/build/cache/backends/), they only support one cache-to
image

The PR linked in the docs is merged and multiple cache-to is also working in local. I'll switch it to same behavior as import cache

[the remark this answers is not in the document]
rutvijmehta-harness: Done

docker.go
```diff
@@ -181,6 +183,12 @@ func (p Plugin) Exec() error {
 		cmds = append(cmds, commandPull(img))
 	}
 
+	// cache export feature is currently not supported for docker driver
+	// hence we have to create docker-container driver
+	if len(p.Build.ExportCache) > 0 {
```
shubham149, line 188: if len(p.Build.ExportCache) > 0 || len(p.Build.ImportCache) > 0 {}

rutvijmehta-harness: Docs do not seem to indicate that docker driver is not supported. It only mentions about the restriction during exports.

https://docs.docker.com/build/cache/backends/

shubham149 (approved):

PR #390, opened 2023-05-15
fix: Use unique build name for build and tag

rutvijmehta-harness:
Bug:

Created 2 dockerfiles where we set the below environment variables and built them in parallel:
`testimg:1.0` : `ENV GOPATH=$HOME/go`
`testimg2:2.0` : `ENV GOPATH=$HOME/go2`

Parallel Step 1:
image

Parallel Step 2:
image

Reproduced the bug where both testimg and testimg2 were the same (since the environment variable set was the same) after they were built. Ideally the should be different:
Commands used:
`testimg:1.0`:
`docker run --name=testcontainer1 rutvijspit/testimg:1.0`
`docker exec testcontainer1 sh -c /usr/bin/env | grep GOPATH`

`testimg2:2.0`:
`docker run --name=testcontainer2 rutvijspit/testimg2:2.0`
`docker exec testcontainer2 sh -c /usr/bin/env | grep GOPATH`

image

After the fix:

Parallel Step 1:
image

Parallel Step 2:
image

Verification that different environement variable is set:
image

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -45,6 +45,7 @@ type (
 	Build struct {
 		Remote      string   // Git remote URL
 		Name        string   // Docker build using default named tag
+		TempTag     string   // Temporary tag used during docker build
 		Dockerfile  string   // Docker build Dockerfile
 		Context     string   // Docker build context
 		Tags        []string // Docker build tags
@@ -195,11 +196,10 @@ func (p Plugin) Exec() error {
 		}
 	}
 
-	buildName := getUniqueBuildName(p.Build.Repo, p.Build.Name)
-	cmds = append(cmds, commandBuild(p.Build, buildName)) // docker build
+	cmds = append(cmds, commandBuild(p.Build)) // docker build
 
 	for _, tag := range p.Build.Tags {
-		cmds = append(cmds, commandTag(p.Build, tag, buildName)) // docker tag
+		cmds = append(cmds, commandTag(p.Build, tag)) // docker tag
 
 		if !p.Dryrun {
 			cmds = append(cmds, commandPush(p.Build, tag)) // docker push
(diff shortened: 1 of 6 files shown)


docker.go
```diff
@@ -582,3 +583,14 @@ func getDigest(buildName string) (string, error) {
 	}
 	return "", errors.New("unable to fetch digest")
 }
+
+func getUniqueBuildName(repo, buildName string) string {
```
shubham149, line 587: What if build is triggered for a run without git clone?

rutvijmehta-harness: Moved away from this approach. We now generate a temporary tag using a random string.

bradrydzewski:
I would recommend using `crypto/rand` to ensure you have a truly random string.  The challenge with `math/rand` is that it is deterministic, and if you seed with time, there is still the possibility of conflicts. 

```Go
import "crypto/rand"

// random generates a random string
func random() string {
	result := make([]rune, 32)
	runes := []rune("0123456789abcdef")
	x := int64(len(runes))
	for i := range result {
		num, err := rand.Int(rand.Reader, big.NewInt(x))
		if err != nil {
			panic("error creating random number") // should never happen
		}
		result[i] = runes[num.Int64()]
	}
	return string(result)
}
```

alternatively you can use a library that handles this. In other Drone projects we use the below library since it is lightweight and feature complete, with zero external dependencies:

```Go
import "github.com/dchest/uniuri"

s := uniuri.New()
```

rutvijmehta-harness:
Thanks @bradrydzewski for the review. I've used `github.com/dchest/uniuri` to generate the temporary tag.

bradrydzewski (approved):

PR #392, opened 2023-06-10
update docker linux amd64/arm64 to 20.10.24

stklcode:
Straight forward update to the latest 2.10.x release, which is 20.10.24 by now.

I personally have a minor build issue that results in unnecessarily larger layers using an older version of the _Docker_ engine. Seems to be resolved with current releases, so I would appreciate to have this resolved in my Drone pipelines as well.


stklcode:
Apparently this PR has been ignored and is not outdated for quite a while. Closing.

PR #395, opened 2023-08-15
Add support for AAD auth for docker-acr

rutvijmehta-harness:
This PR adds support for AAD auth for docker-acr. The previous implementation assumes admin credentials and uses them as username and password. However that may not be the case always.

The new implementation uses `clientId`, `clientSecret` / `clientCert`, `tenantId` to get the ACR token which can be used as password along with the default username `00000000-0000-0000-0000-000000000000`. If `subscriptionId` is provided, it also retrieves a url to view the artifact. If not provided, the registry url is used.

Screenshot of how the artifact would look using the above url:
image
image


cmd/drone-docker/main.go
```diff
@@ -112,6 +112,12 @@ func main() {
 			Usage:  "don't start the docker daemon",
 			EnvVar: "PLUGIN_DAEMON_OFF",
 		},
+		cli.StringFlag{
+			Name:   "daemon.registry",
+			Usage:  "daemon registry",
+			Value:  "https://index.docker.io/v1/",
+			EnvVar: "DAEMON_REGISTRY,PLUGIN_REGISTRY,DOCKER_REGISTRY",
```
rutvijmehta-harness, line 119: For non-acr builds, `DAEMON_REGISTRY` will not be set and we will use `PLUGIN_REGISTRY`

rutvijmehta-harness:
@tphoney we might have to update the go version (1.18+) for the azure sdk to work.

`go vet ./...` is failing in drone (go version 1.17) but passing on local (go version 1.20)

Ref:
1. https://github.com/Azure/azure-sdk-for-go/issues/17472
2. https://github.com/google/go-cloud/issues/3141#issue-1291574283

Let me know if that's okay. I'll update the go version in drone yml.

tphoney:
@rutvijmehta-harness can you bump the go version, as you suggested in the go.mod file and the .drone.yml file in a separate commit . 
I will look at the test of the pr now

cmd/drone-docker/main.go
tphoney: i am not particularly happy with this, as we are changing the default behaviour for the "docker.registry" why dont we just use the env variable PLUGIN_REGISTRY like we do for the other providers.

or is DAEMON_REGISTRY something totally new ?

tphoney (changes requested):

cmd/drone-docker/main.go
rutvijmehta-harness: We have a hard assumption that `docker.registry` will point to a page where we can view the artifact. This is not the case for ACR. We get the url from Azure using the API implemented in the PR.

For kaniko, we have a separate struct for artifact, which is why it is straight forward there.

Ref:
1. https://github.com/drone/drone-kaniko/blob/main/kaniko.go#L239
2. https://github.com/drone/drone-kaniko/blob/main/cmd/kaniko-acr/main.go#L261

rutvijmehta-harness: We cannot update `PLUGIN_REGISTRY` as the push still needs to go to the input registry.

rutvijmehta-harness: I'm okay with adding `ArtifactRegistry` field to `docker.Daemon` and use it while writing the digest.

`drone.WritePluginArtifactFile(p.Daemon.RegistryType, p.ArtifactFile, p.Daemon.ArtifactRegistry, p.Build.Repo, digest, p.Build.Tags)`

tphoney: i was thinking of keeping the changes to main.go as https://github.com/drone-plugins/drone-docker/compare/acr...tphoney:drone-docker:acr?expand=1 and then adding logic into acr for what ever is needed. 
changing the handling inside of main.go and overriding what is standard is dangerous.
inside of the acr we can change as much as necessary.

rutvijmehta-harness: Moved to use `p.Daemon.ArtifactRegistry` which is a different path and not standard: https://github.com/drone-plugins/drone-docker/pull/395/commits/c64f19dec75948ec28095e4fc6e9ff5f12f2437e

Please approve and merge if it looks okay.

cmd/drone-acr/main.go
```diff
@@ -62,3 +263,9 @@ func getenv(key ...string) (s string) {
 	}
 	return
 }
+
+type strct struct {
```
eoinmcafee00, line 267: Naming? What exactly is this?

rutvijmehta-harness: This code was taken from `drone-kaniko`: https://github.com/drone/drone-kaniko/blob/main/cmd/kaniko-acr/main.go#L442

Once the implementation is approved, I can make this change and take approval from US team and merge this PR.

tphoney (changes requested):
This is still not good enough. 
my objection is to the changes in the drone-docker/main.go adding artifact.registry ```EnvVar: "ARTIFACT_REGISTRY,PLUGIN_REGISTRY,DOCKER_REGISTRY",``` make it EnvVar: "ARTIFACT_REGISTRY". What would happen if a user passes those 3 env variables ?

the changing of docker.go 
```fmt.Printf("Got artifact reg %s and digest %s\n", p.Daemon.ArtifactRegistry, digest)``` will only work for ACR, not other docker registrys. thus printing out broken information for gcr/docker ...

if you want to make these kinds of changes to core parts of the docker plugin. You will need to test against docker / acr / gcr / ecr / heroku.

please only make additive changes to drone-docker and docker.go. Additive changes are fine. 
but what you are doing now only works for acr, and may have unforseen affects on the other registry types.

rutvijmehta-harness:
> This is still not good enough. my objection is to the changes in the drone-docker/main.go adding artifact.registry `EnvVar: "ARTIFACT_REGISTRY,PLUGIN_REGISTRY,DOCKER_REGISTRY",` make it EnvVar: "ARTIFACT_REGISTRY". What would happen if a user passes those 3 env variables ?

`ARTIFACT_REGISTRY` would take precedence. Why would user set this to the env if `drone-docker` does not support it? This will be true for any new env variables we add.

> the changing of docker.go `fmt.Printf("Got artifact reg %s and digest %s\n", p.Daemon.ArtifactRegistry, digest)` will only work for ACR, not other docker registrys. thus printing out broken information for gcr/docker ...

This was only for debugging. I'll remove it.

> if you want to make these kinds of changes to core parts of the docker plugin. You will need to test against docker / acr / gcr / ecr / heroku.

I've already tested against docker. We have the fallback env variable `PLUGIN_REGISTRY` which will take care of any regression.

> please only make additive changes to drone-docker and docker.go. Additive changes are fine. but what you are doing now only works for acr, and may have unforseen affects on the other registry types.

Why do you think it only works for acr?

tphoney (approved):

PR #410, opened 2023-10-04
Remove .git suffix from opencontainer source label

chamburr:
The `org.opencontainers.image.source` label currently look like this: `https://github.com/octocat/hello-world.git`. Looking through GitHub however, I believe that the .git suffix should not be there. Furthermore, GitHub packages support automatically linking an uploaded container image to the source repository using labels, but this feature does not work with the extra .git suffix. This pull request therefore removes the .git suffix from the label.

References:
https://github.com/search?q=org.opencontainers.image.source&type=code
https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images


chamburr:
@tphoney Do you mind taking a look? Thank you

PR #412, opened 2023-10-19
ci-3011 - workload identity/OIDC support for GAR/GCR

eoinmcafee00:
Normal service account flow:
Screenshot 2023-10-19 at 14 18 43

New OIDC Flow:
Screenshot 2023-10-19 at 14 06 25


README.md
```diff
@@ -113,6 +113,46 @@ docker run --rm \
...
+type: docker
+
+steps:
+  - name: push-to-gar
+    image: plugins/gcr
+    pull: never
+    settings:
+      tag: latest
+      repo: project-id/repo/image-name
+      registry_type: GAR
```
tphoney, line 130: this should go into plugins.drone.io as well

docker.go
```diff
@@ -34,11 +34,12 @@ type (
...
-		Username string // Docker registry username
-		Password string // Docker registry password
-		Email    string // Docker registry email
-		Config   string // Docker Auth Config
+		Registry    string // Docker registry address
+		Username    string // Docker registry username
+		Password    string // Docker registry password
+		Email       string // Docker registry email
+		Config      string // Docker Auth Config
+		AccessToken string // OIDC Access Token
```
tphoney, line 42: can we be consistent with naming, and add oidc to the variable name

docker.go
```diff
@@ -270,6 +283,17 @@ func commandLogin(login Login) *exec.Cmd {
 	)
 }
 
+func commandLoginOIDC(login Login, accessToken string) *exec.Cmd {
```
tphoney, line 286: add accesstoken to the func name, and oidc to the accesstoken varible name

tphoney (changes requested):

docker.go
eoinmcafee00: I kept it generic as access token because that's what it is... we might get an access token via a different flow in AWS?

README.md
eoinmcafee00: will do.

docker.go
eoinmcafee00: renamed

PR #413, opened 2023-10-19
ci-9739 - workload identity/OIDC support for GAR/GCR 

eoinmcafee00:


cmd/drone-gcr/main.go
```diff
@@ -73,24 +105,24 @@ func loadConfig() Config {
...
 }
 
 func main() {
 	config := loadConfig()
+	if config.AccessToken != "" {
+		os.Setenv("ACCESS_TOKEN", config.AccessToken)
+	} else if config.Username != "" && config.Password != "" {
+		os.Setenv("DOCKER_USERNAME", config.Username)
+		os.Setenv("DOCKER_PASSWORD", config.Password)
+		os.Setenv("", strconv.FormatBool(config.WorkloadIdentity))
```
vistaarjuneja, line 121: is this expected? what will setting an empty key do?

vistaarjuneja (approved):

raghavharness (approved):

bradrydzewski:
@eoinmcafee00 I would recommend splitting gcr and gar into two separate docker images (add `cmd/drone-gar`).  The reason is that they are two separate products.  For this reason, when an end-user searches for "gar" in the marketplace it may not be intuitive that they should use "gcr" with "gar" as the type

eoinmcafee00:
Hey @bradrydzewski 
PR for this: https://github.com/drone-plugins/drone-docker/pull/415

PR #445, opened 2024-06-17
base image connector for vm

Aishwarya-Lad:
Adding provision to allow the base image to be pulled from a private docker registry.
The screen shots show pulling from private repo - docker.io/**************/test-private:1.0.1

Screenshot 2024-06-18 at 12 47 00 AM

Screenshot 2024-06-18 at 12 46 49 AM

### DIFF
--- a/cmd/drone-docker/main.go
+++ b/cmd/drone-docker/main.go
@@ -223,6 +223,21 @@ func main() {
 			EnvVar: "PLUGIN_PASSWORD,DOCKER_PASSWORD",
 		},
 		cli.StringFlag{
+			Name:   "docker.baseimageusername",
+			Usage:  "Docker username for base image registry",
+			EnvVar: "PLUGIN_DOCKER_USERNAME,PLUGIN_BASE_IMAGE_USERNAME",
+		},
+		cli.StringFlag{
+			Name:   "docker.baseimagepassword",
+			Usage:  "Docker password for base image registry",
+			EnvVar: "PLUGIN_DOCKER_PASSWORD,PLUGIN_BASE_IMAGE_PASSWORD",
+		},
+		cli.StringFlag{
+			Name:   "docker.baseimageregistry",
+			Usage:  "Docker registry for base image registry",
+			EnvVar: "PLUGIN_DOCKER_REGISTRY,PLUGIN_BASE_IMAGE_REGISTRY",
+		},
+		cli.StringFlag{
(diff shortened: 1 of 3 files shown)


bradrydzewski:
We have the existing `DOCKER_PLUGIN_CONFIG` options which allows you to pass the contents of a `dockerconfig.json` with additional credentials.  It is more flexible than providing base credentials because you can provide multiple credentials, in the event that you have multiple base images in multiple registries.
https://github.com/drone-plugins/drone-docker/blob/master/cmd/drone-docker/main.go#L233C13-L233C26

We should be able to use this existing, more flexible option instead of adding a new options.

cc @jimsheldon if you have additional questions

bradrydzewski (changes requested):
instead of adding new variables, leverage `DOCKER_PLUGIN_CONFIG` for passing one or more registry credentials.

Aishwarya-Lad:
@bradrydzewski the reason we decided with this approach was because, docker would not support login to 2 private repositories that have same registry url. 
So we decided with docker login  PLUGIN_DOCKER_REGISTRY -> build image -> docker login PLUGIN_REGISTRY. If we were to pass the DOCKER_PLUGIN_CONFIG, we might still need to use PLUGIN_DOCKER_REGISTRY. 

https://github.com/moby/moby/issues/37569

cc. jamie-harness

Aishwarya-Lad:
> We have the existing `DOCKER_PLUGIN_CONFIG` options which allows you to pass the contents of a `dockerconfig.json` with `auth` credentials. It is more flexible than providing base credentials because you can provide multiple credentials, in the event that you have multiple base images in multiple registries. https://github.com/drone-plugins/drone-docker/blob/master/cmd/drone-docker/main.go#L233C13-L233C26
> 
> We should be able to use the existing `DOCKER_PLUGIN_CONFIG` option instead of adding new parameters, given it is much more flexible. I recommend we close this pull request and use `DOCKER_PLUGIN_CONFIG`


@bradrydzewski as confirmed in the slack thread, we are moving ahead with the approach of writing config.json in drone plugins from variables `PLUGIN_DOCKER_USERNAME` , `PLUGIN_DOCKER_PASSWORD` and `PLUGIN_DOCKER_REGISTRY` instead of passing `DOCKER_PLUGIN_CONFIG` from CI side. It is not straightforward to decrypt the secrets with out current setup, as mentioned in detail in the slack thread.

PR #446, opened 2024-06-23
update docker config 

Aishwarya-Lad:


docker.go
```diff
@@ -154,10 +158,37 @@ func (p Plugin) Exec() error {
 		os.MkdirAll(dockerHome, 0600)
 
 		path := filepath.Join(dockerHome, "config.json")
-		err := os.WriteFile(path, []byte(p.Login.Config), 0600)
+		file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
```
jimsheldon, line 161: Can you help me understand why we need this change? If it is needed, the following error message is not accurate, since it says "Error writing" where this is just a `OpenFile` operation.

docker.go
```diff
@@ -270,6 +301,35 @@ func (p Plugin) Exec() error {
 	return nil
 }
 
+// helper function to set the credentials
+func setDockerAuth(username, password, registry, baseImageUsername,
+	baseImagePassword, baseImageRegistry string) ([]byte, error) {
+	var credentials []docker.RegistryCredentials
+	// add only docker registry to the config
+	if password != "" && strings.Contains(registry, "docker") {
+		dockerConfig := docker.NewConfig()
```
jimsheldon, line 310: `dockerConfig` must first be first defined outside of the `if` condition. Currently this causes an error
```
./docker.go:310:3: dockerConfig declared and not used
```

docker.go
```diff
@@ -270,6 +301,35 @@ func (p Plugin) Exec() error {
...
+		pushToRegistryCreds := docker.RegistryCredentials{
+			Registry: registry,
+			Username: username,
+			Password: password,
+		}
+		// push registry auth
+		credentials := append(credentials, pushToRegistryCreds)
```
jimsheldon, line 317: Same as above, `credentials` must be defined first outside the `if` condition
```
./docker.go:317:3: credentials declared and not used
```

docker.go
```diff
@@ -270,6 +301,35 @@ func (p Plugin) Exec() error {
...
+		pullFromRegistryCreds := docker.RegistryCredentials{
+			Registry: baseImageRegistry,
+			Username: baseImageUsername,
+			Password: baseImagePassword,
+		}
+		// base image registry auth
+		credentials = append(credentials, pullFromRegistryCreds)
+	}
+	// Creates docker config for both the registries used for authentication
+	return dockerConfig.CreateDockerConfigJson(credentials)
```
jimsheldon, line 330: Due to the above, this fails
```
./docker.go:330:9: undefined: dockerConfig
```

jimsheldon (changes requested):

docker.go
Aishwarya-Lad: We need this change to be able to append to the config file in case it already exists. I kept the error message generalized but I can fix it to be more specific.

docker.go
Aishwarya-Lad: Yea, the latest commit hasn't been pushed. This is already addressed.

docker.go
jimsheldon: Thanks, the following error is probably sufficient in this case

internal/docker/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+	jsonBytes, err := json.Marshal(c)
+	if err != nil {
+		return nil, errors.New("failed to serialize docker config json")
+	}
+
+	return jsonBytes, nil
+}
+
+func WriteDockerConfig(data []byte, path string) (string error) {
```
rutvijmehta-harness, line 75: Where is this used?

Aishwarya-Lad: I created this helper function to write, but ended up using go's utility directly.

internal/docker/config.go
```diff
@@ -0,0 +1,89 @@
...
+	c.Auths[registry] = Auth{Auth: encodedString}
+}
+
+func (c *Config) SetCredHelper(registry, helper string) {
+	c.CredHelpers[registry] = helper
+}
+
+func (c *Config) CreateDockerConfigJson(credentials []RegistryCredentials) ([]byte, error) {
+	for _, cred := range credentials {
+		if cred.Registry != "" && strings.Contains(cred.Registry, "docker") {
```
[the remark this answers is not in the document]
Aishwarya-Lad: Yea, you are correct. I wanted to make this function call exclusive to docker step but then jfrog and other docker supported artifacts will not be served.

jamie-harness (approved):

PR #449, opened 2024-07-12
Change docker login method, from config writing to login func

Aishwarya-Lad:
This upgrade includes changes in base buildx version:

Changed the login to docker login command instead of using the Config struct for OIDC changes.
Made the decision to use login instead of writing file directly.
Why the change was made - The login function integrates with various credential helpers, it also handles different registry specific logic in a better way, as opposed to config write where different registries need to be addressed differently.
It handles any changes in the authentication process across different Docker versions.

https://docs.google.com/document/d/153BXzvNWUZ-XD2qOIuPDQ3XWVSI89tR0uBppLNS-F-4/edit

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -158,34 +158,33 @@ func (p Plugin) Exec() error {
 		os.MkdirAll(dockerHome, 0600)
 
 		path := filepath.Join(dockerHome, "config.json")
-		file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+		err := os.WriteFile(path, []byte(p.Login.Config), 0600)
 		if err != nil {
 			return fmt.Errorf("Error writing config.json: %s", err)
 		}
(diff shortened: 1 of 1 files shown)


docker.go
```diff
@@ -158,34 +158,33 @@ func (p Plugin) Exec() error {
...
 			return fmt.Errorf("Error writing config.json: %s", err)
 		}
-		err = os.WriteFile(path, []byte(p.Login.Config), 0600)
-		if err != nil {
-			return fmt.Errorf("Error writing config.json: %s", err)
-		}
-		file.Close()
 	}
 
 	// add base image docker credentials to the existing config file, else create new
```
hemanthmantri, line 167: Shouldn't this comment be removed now that we're not adding credentials?

Aishwarya-Lad: removed it now, originally decided to keep the comments because we discussed to capture all the changed approach decisions in the comments. But it makes sense to remove it too I guess.

PR #461, opened 2024-11-27
Docker saving tarball and no-push support.

Anshika2203:
DRAFT PR - NOT FINAL


bradrydzewski:
this plugin already has support for disabling push using the `dry_run` flag
> boolean if the docker image should not be pushed at the end

see https://plugins.drone.io/plugins/docker

Anshika2203:
> this plugin already has support for disabling push using the `dry_run` flag
> 
> > boolean if the docker image should not be pushed at the end
> 
> see https://plugins.drone.io/plugins/docker

oh I will look into it.

PR #470, opened 2025-02-05
Do not pre-pull cache images

gzm0:
Since 20.18.6 / 2140a2be40e8a5511051ffb38974ef83064b49a6 / #462, buildkit is always enabled.

Therefore, there is no need anymore to pre-pull the cached images, `--cache-from` will cause buildkit to pull as needed.

This is a slimmed down alternative to #360.


neil-ach (approved):
Nice.

PR #472, opened 2025-02-12
feat: [CI-16181]: New input skip_push_if_tag_exists

Ompragash:
New input `skip_push_if_tag_exists` flag to the Drone ECR plugin, allowing users to skip the image push if the specified tag already exists in the repository. The logic supports both mutable and immutable repositories, checking for existing tags before proceeding with the build and push.

### DIFF
--- a/cmd/drone-ecr/main.go
+++ b/cmd/drone-ecr/main.go
@@ -22,6 +22,10 @@ import (
 	docker "github.com/drone-plugins/drone-docker"
 )
 
+type ecrAPI interface {
+	DescribeImages(*ecr.DescribeImagesInput) (*ecr.DescribeImagesOutput, error)
+}
+
 const defaultRegion = "us-east-1"
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/drone-ecr/main.go
```diff
@@ -114,6 +119,34 @@ func main() {
 	os.Setenv("DOCKER_PASSWORD", password)
 	os.Setenv("PLUGIN_REGISTRY_TYPE", "ECR")
 
+	// Skip if tag already exits for both mutable and immutable repos
+	if skipPushIfTagExists {
+		tagInput := getenv("PLUGIN_TAG", "PLUGIN_TAGS")
```
jamie-harness, line 124: Can you use the variable parsed by golang tool instead of reading directly from env?

jamie-harness: nvm ignore my comment

jamie-harness (approved):

PR #494, opened 2025-07-31
feat: [CI-18308]: Add Cosign Image Signing Support

Ompragash:
Adds container image signing capability using Cosign directly within the drone-docker plugin. Now supports signing images using their immutable digests with proper fallback to tag-based signing.

New Inputs:
- PLUGIN_COSIGN_PRIVATE_KEY: Private key content (PEM) or file path
- PLUGIN_COSIGN_PASSWORD: Optional password for encrypted keys
- PLUGIN_COSIGN_PARAMS: Optional additional cosign parameters

Features:
- Signs using immutable digest references (repo@sha256:digest) for better security
- Falls back to tag-based signing if digest unavailable
- Non-interactive operation with COSIGN_YES=true
- Handles both embedded PEM keys and file paths
- Signs after push completes, ensuring registry digest is used

Testing:

https://github.com/user-attachments/assets/ea2383a3-0a53-4af3-8208-e34a128c9b9d

### DIFF
--- a/docker/docker/Dockerfile.linux.amd64
+++ b/docker/docker/Dockerfile.linux.amd64
@@ -3,7 +3,7 @@ FROM docker:28.1.1-dind
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 
 # Install cosign for container image signing
-RUN wget -O /usr/local/bin/cosign https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-amd64 \
+RUN wget -O /usr/local/bin/cosign https://github.com/sigstore/cosign/releases/download/v2.5.3/cosign-linux-amd64 \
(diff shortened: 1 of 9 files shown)


docker/docker/Dockerfile.linux.amd64
```diff
@@ -2,5 +2,9 @@ FROM docker:28.1.1-dind
 
 ENV DOCKER_HOST=unix:///var/run/docker.sock
 
+# Install cosign for container image signing
+RUN wget -O /usr/local/bin/cosign https://github.com/sigstore/cosign/releases/latest/download/cosign-linux-amd64 \
```
jamie-harness, line 6: Do we have a way to version this in case of needing to change version or debug?

PR #514, opened 2026-07-07
feat: [CI-23609]: docker files for windows 2025

tapankarangiya:
Screenshot 2026-07-08 at 3 14 32 PM
built drone-docker windows 2025 and tested it out

### DIFF
--- a/docker/acr/Dockerfile.windows.amd64.ltsc2025
+++ b/docker/acr/Dockerfile.windows.amd64.ltsc2025
@@ -1,0 +1,10 @@
+# escape=`
+FROM plugins/docker:windows-ltsc2025-amd64
+
+LABEL maintainer="Drone.IO Community <drone-dev@googlegroups.com>" `
+  org.label-schema.name="Drone ACR" `
+  org.label-schema.vendor="Drone.IO Community" `
+  org.label-schema.schema-version="1.0"
+
+ADD release/windows/amd64/drone-acr.exe C:/bin/drone-acr.exe
+ENTRYPOINT [ "C:\\bin\\drone-acr.exe" ]
--- a/docker/acr/manifest.tmpl
+++ b/docker/acr/manifest.tmpl
@@ -29,3 +29,9 @@ manifests:
       architecture: amd64
       os: windows
       version: ltsc2022
+  -
+    image: plugins/acr:{{#if build.tag}}{{trimPrefix "v" build.tag}}-{{/if}}windows-ltsc2025-amd64
+    platform:
+      architecture: amd64
+      os: windows
+      version: ltsc2025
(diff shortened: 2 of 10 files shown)


raghavharness (changes requested):
## Summary

This PR adds Windows Server LTSC 2025 (`nanoserver:ltsc2025`) Docker build support for the drone-docker plugin family. It introduces new `Dockerfile.windows.amd64.ltsc2025` files for acr/docker/ecr/gar/gcr and appends a `windows-ltsc2025-amd64` entry to each service's `manifest.tmpl`. The base `docker` image pulls cosign + docker.exe (via innoextract) and notably skips the `netapi32.dll` copy since ltsc2025 nanoserver ships it natively.

## Verdict

**Concerns** — see below.

## Concerns

- **Risk:** No CI pipeline actually builds these new Dockerfiles. `.drone.yml` only defines `windows-1809` and `windows-ltsc2022` pipelines, and `.harness/harness.yaml` hardcodes `Dockerfile.windows.amd64.ltsc2022`. Without a new `windows-ltsc2025` pipeline (mirroring the ltsc2022 block), the `windows-ltsc2025-amd64` images referenced in the manifests will never be pushed — and `plugins/manifest` runs with `ignore_missing: true`, so this fails silently rather than erroring. Please add the build steps or confirm they land in a follow-up.
- **Question:** `DOCKER_VERSION 19.03.1` and DockerToolbox are quite old (matching the 1809/ltsc2022 files). Intentional to stay consistent, or worth bumping for a new-OS image?

raghavharness (approved):
## Summary

This PR adds Windows Server 2025 (`ltsc2025`) support to the drone-docker plugin family by introducing new `Dockerfile.windows.amd64.ltsc2025` files for `acr`, `docker`, `ecr`, `gar`, and `gcr`, and appending matching `ltsc2025-amd64` entries to each `manifest.tmpl`. The base `docker` image is modernized: it pulls the standalone Docker CLI 28.4.0 from download.docker.com (replacing the old DockerToolbox/innoextract path) and relies on `nanoserver:ltsc2025` shipping `netapi32.dll` natively, so the servercore DLL copy is dropped. The satellite plugins (`acr`/`ecr`/`gar`/`gcr`) simply `FROM plugins/docker:windows-ltsc2025-amd64`.

## Verdict

**Concerns** — see below.

## Concerns

- **Risk:** `.drone.yml` has no `windows-ltsc2025` build pipeline. The existing jobs only build `1809` and `ltsc2022` (e.g. `windows-ltsc2022` uses `pool: windows-2022`), so nothing ever builds or pushes the `plugins/docker:windows-ltsc2025-amd64` image the new Dockerfiles and manifests reference. The satellite Dockerfiles will fail to pull that base, and the new manifest entries will point at images that are never published. A `windows-ltsc2025` pipeline (with a Server 2025 pool) plus wiring into the `notifications-*` `depends_on` appears to be missing from this PR.
- **Question:** Was the `ltsc2025` build verified through CI, or only locally as the screenshot suggests? Without the pipeline addition, the manifest step relies on `ignore_missing: true` and will silently omit ltsc2025.

devkimittal (approved):

Gargi-thakur01 (approved):

Gargithakur01 (approved):

